PR #11, opened 2021-01-16
completed literature search tests

JNKHunter:


src/test/resources/queries/lit-search-3.sparql
```diff
@@ -0,0 +1,49 @@
...
+PREFIX  litp:  <http://lit-search-api/property/>
+PREFIX  lita:  <http://lit-search-api/attribute/>
+
+CONSTRUCT
+{
+    ?detent litp:entityLink _:el .
+    _:el lita:confidence ?linkconf .
+    ?linkconf a xsd:float .
+    _:el litp:entityLink ?concept .
+    #?detent litp:entityLink ?concept .
```
puremachinery, line 19: Is it valuable to keep this commented line? If so, maybe add another line explaining when one would incorporate it back in.

src/test/resources/queries/lit-search-4.sparql
```diff
@@ -0,0 +1,60 @@
+PREFIX  schema: <http://schema.org/>
```
puremachinery, line 1: General comment for lit search tests: Did we generate the queries or were they provided users who run searches? Are we confident they are representative of all relevant literature search use cases? (I expect so, but seems worth a quick check.)

puremachinery: Nevermind, I see that these just map to the queries at https://mygithub.gsk.com/gsk-tech/kg-literature-search-api/tree/master/graphbuild/queries

src/test/resources/queries/lit-search-3.sparql
puremachinery: Nevermind, I see that the query comment is present in https://mygithub.gsk.com/gsk-tech/kg-literature-search-api/tree/master/graphbuild/queries query #3 so this question applies over there more than here.

puremachinery (approved):
Looks straightforward.

PR #14, opened 2021-01-21
Add Construct query case to current ADT model

shihch:


src/main/scala/QueryConstruct.scala
```diff
@@ -0,0 +1,80 @@
...
+      toStringVal(triple.getSubject),
+      toStringVal(triple.getPredicate),
+      toStringVal(triple.getObject)
+    )
+  }
+
+
+
+  def main(args: Array[String]): Unit = {
+//    val q = Source.fromFile("src/test/resources/queries/q3-union.sparql").mkString
```
puremachinery, line 75: Any reason not to remove commented code?

puremachinery (changes requested):
Remove commented line of code, otherwise seems good.

Other note: John recently checked in scalastyle - might be good to run that over this, but not required for approval.

src/main/scala/StringLike.scala
```diff
@@ -14,5 +14,6 @@ object StringVal {
   final case class STRING(s:String) extends StringVal
   final case class VARIABLE(v:String) extends StringVal
   final case class URIVAL(u:String) extends StringVal
+  final case class BLANK(ref:String) extends StringVal
```
JNKHunter, line 17: Sweet! Was going to add Blank node support :)

JNKHunter (changes requested):
Looks good Clark. There were some Scalastyle warnings (You can look at the 'checks' tab). If we could clear those out, I'm good to merge. We'll need to figure out how to have checks fail when scalastyle has warnings.

src/main/scala/QueryConstruct.scala
shihch: The main method is removed, which is no need for now.

shihch:
scala style check is now passed.

puremachinery (approved):
Looks like this covers the comment and the scalastyle changes. Great.

JNKHunter (approved):

PR #20, opened 2021-01-25
Added tests for parsing large sparql query

JNKHunter:
Also added support and tests for parsing greater than and less than filters


shihch (approved):
LGTM

PR #23, opened 2021-01-29
Apply visitor pattern to operate on parsed sparql objects, add a simple implementation to test

shihch:


src/main/scala/com/gsk/kg/sparql/impl/ExprToText.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.gsk.kg.sparqlparser.StringVal._
+import com.gsk.kg.sparqlparser.{FilterFunction, StringFunc, StringLike, StringVal}
+
+object ExprToText {
+
+  implicit class StringLikeToText(st: StringLike) {
+    def text: String = {
+      st match {
+        case sv: StringVal =>
+          sv match {
```
JNKHunter, line 14: As we build our impl, let's look into Jena constructs like [Node](https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/Node.html) and [Triple](https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/Triple.html) also [Node Factory](https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/NodeFactory.html)

JNKHunter (approved):
Looks good to me. See inline comment for future considerations.

PR #24, opened 2021-02-02
downgrade to scala 2.12

shihch:


shihch:
Cancel this pull request and create another one by adding cross compile.

PR #28, opened 2021-02-05
Merge spark engine

pepegar:
This PR merges the Spark engine into the algebra parser.

## modules

We'll have two different modules inside this project now: `bellman-algebra-parser` in the `modules/parser` directory, and `bellman-spark-engine` in the `modules/engine` directory.

We'll also have a _synthetic_ module that aggregates both, so that we can do things such as `sbt +test` and it runs tests for both.

## Actions

While merging the code from the spark engine I've also brought some files that I think can help with the maintenance of the project:

### dependabot

Dependabot will create PRs when new versions of actions are created.

### scala-steward

Scala Steward checks maven central and creates PRs in the repo when it detects a new version of a library.

### release

The release action is executed on every commit on master and releases versions to sonatype.  For normal commits it generates a SNAPSHOT version with the commit hash appended (see https://oss.sonatype.org/index.html#nexus-search;quick~bellman-spark-engine)

When we push a tag with the `v*.*.*` shape to the repo, it will generate a release that will end up in maven central.

## Secrets

Since our CI runs in Github actions, we've introduced some secrets to the build, anyone with admin privs can check them.  They're required for the `release.yml` github action.


JNKHunter (approved):
LGTM.

shihch (approved):
Excellent! Thanks Pepe.

PR #34, opened 2021-02-10
add query samples and parser tests.

shihch:
add limit and offset query support in parser.
add distinct query support.
add function isBlank and negate (!) support.

### DIFF
--- a/modules/parser/src/test/scala/com/gsk/kg/sparqlparser/QuerySamplesTestSpec.scala
+++ b/modules/parser/src/test/scala/com/gsk/kg/sparqlparser/QuerySamplesTestSpec.scala
@@ -152,20 +152,19 @@ class QuerySamplesTestSpec extends AnyFlatSpec {
     }
   }
 
-  // TODO: Comment out for now, need to work on parser since when BIND in different position Jena generates different representation
-//  "Test BIND in another position in the query" should "parse to same as q12" in {
-//    val query = QuerySamples.q13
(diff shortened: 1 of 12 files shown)


modules/parser/src/test/scala/com/gsk/kg/sparqlparser/QuerySamplesTestSpec.scala
```diff
@@ -0,0 +1,185 @@
...
+//    println(expr)
+//    expr match {
+//      case Select(vs, Filter(funcs,Extend(to,from,BGP(_)))) =>
+//        assert(vs.nonEmpty && vs.size == 3)
+//        assert(funcs.size == 1)
+//        assert(to == VARIABLE("?o"))
+//      case _ =>
+//        fail
+//    }
+//  }
```
pepegar, line 168: Maybe we can leave the test as ignored then?

```suggestion
  "Test BIND in another position in the query" should "parse to same as q12" ignore {
    val query = QuerySamples.q13
    val expr = QueryConstruct.parseADT(query)
    println(expr)
    expr match {
      case Select(vs, Filter(funcs,Extend(to,from,BGP(_)))) =>
        assert(vs.nonEmpty && vs.size == 3)
        assert(funcs.size == 1)
        assert(to == VARIABLE("?o"))
      case _ =>
        fail
    }
  }
```

pepegar (approved):
Added a small comment, but looks good otherwise!  Feel free to merge either way :)

PR #39, opened 2021-02-12
Sparql parser tests on sample queries

shihch:
Factor out Query types (Construct, Select, Describe, Ask) from Expr
Change Select in Expr to Project
Make FilterFunction and StringLike all are Expression trait
add support of ||, && binary expressions
add support of replace function
Two of sample queries failed to parse, put under ignore and will fix them later.


pepegar (approved):
:+1: Looks good to me!

PR #41, opened 2021-02-12
Create basic structure for the spark Engine

pepegar:
This PR introduces the basic structure for the Spark engine.100%

first of all, I've need to juggle with dependencies a bit to make both
Jena and Spark coexist in the project.  Jena uses a newer version of
JAckson, while spark uses an old one.  I needed to downgrade Jena's.

## Engine

The entry point for the project is the `Engine` object.  It contains
the evaluate method that will receive an `Expr`, a `DataFrame`, and
will apply the SparQL algebra expression tothe DF.100%

So far, only `SELECT` with Basic Graph Patterns are supported.  I
still have to add some more tests to it to check that the semantics
are correct.

## Multiset

The other central piece for the engine is the `Multiset` datatype.
I've modeled BGP & Select semantics described on
https://www.w3.org/2001/sw/DataAccess/rq23/rq24-algebra.html directly
there.

One can think of `Multiset` as a DataFrame with some bindings, that
can be joined with other `MultiSet`s, or one can project a bunch of
variables from it using the `select` method.

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
@@ -8,6 +8,11 @@ final case class Multiset(
   dataframe: DataFrame
 ) {
 
-  def join(other: Multiset) = this
+  def join(other: Multiset) = (this, other) match {
+    case (a, b) if a.isEmpty => b
+    case (a, b) if b.isEmpty => a
+  }
+
+  def isEmpty: Boolean = bindings.isEmpty && dataframe.isEmpty
 
 }
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
@@ -1,0 +1,13 @@
+package com.gsk.kg.engine
+
+import com.gsk.kg.sparqlparser.StringVal.VARIABLE
+import org.apache.spark.sql.DataFrame
+
+final case class Multiset(
+  bindings: Set[VARIABLE],
+  dataframe: DataFrame
+) {
+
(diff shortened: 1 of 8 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -0,0 +1,65 @@
...
+    case (Multiset(aBindings, aDF), Multiset(bBindings, bDF)) if aBindings.intersect(bBindings).isEmpty =>
+      val df = aDF.as("a")
+        .crossJoin(bDF.as("b"))
+      Multiset(aBindings.union(bBindings), df)
+    case (a, b) =>
+      val common = a.bindings.intersect(b.bindings)
+      val df = a.dataframe.as("a")
+        .join(b.dataframe.as("b"), common.map(_.s).toSeq)
+      Multiset(a.bindings.union(b.bindings), df)
+  }
```
pepegar, line 43: i'd really appreciate someone to double check this piece.

I'm pretty sure of the join when there are common vars, but would like another pair of eyes to for the crossJoin piece.

shihch: Yes, it is the cross join.
For optimization concern, do we want to perform cross joins after all the join operations are done?
Also we may start the filtering from triples without/fewer variables which cuts down data load from the data source. Not sure if Spark does this in its optimizer.

shihch (approved):

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
pepegar: Yeah, I think both optimizations make sense.

I've created #44 and #45 to keep track of them

PR #46, opened 2021-02-16
Complete query parser on sample queries

shihch:
Query parser successfully test on 34 cases out of 35 queries.
One exception is the BIND statement before BGP, which generates
```
join( extend(((var, expression)), (table unit)), bgp(..))
```
The (table unit) seems specific to jena: 
https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/sparql/algebra/table/TableUnit.html

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,7 @@ lazy val Versions = Map(
   "droste"             -> "0.8.0",
   "spark"              -> "2.4.7",
   "spark-testing-base" -> "2.4.5_0.14.0",
-  "jackson"            -> "2.6.7",
+  "jackson"            -> "2.12.1",
 )
 
 inThisBuild(List(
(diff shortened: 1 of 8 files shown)


pepegar (approved):
Looks good to me.

Regarding Jena's specific `table unit`, what can we do about that?  is piece of the algebra gonna get generated by any of our queries?

JNKHunter:
> Looks good to me.
> 
> Regarding Jena's specific `table unit`, what can we do about that? is piece of the algebra gonna get generated by any of our queries?

This query is based on a real world scenario, and we must produce the correct query results.

We can safely ignore the `table unit` construct if Jena is only using it for internal operation. Otherwise if we are dependent on the construct we should parse `table unit` to something we can use.

PR #47, opened 2021-02-16
Evaluate union queries

pepegar:
This PR makes it possible to evaluate union queries.

## Evaluating UNION queries

We rely on Spark DataSet's `union` functions, and depending on the
shared bindings of the different multisets we keep the same DF
structure, or add the needed columns.

## `sparql` interpolator

I also introduced some syntax sugar for making it possible to parse
sparql queries directly, using the `sparql` interpolator.

``` scala
val query: Query = sparql"""
  SELECT {?s ?p ?o}
  WHERE {?s ?p ?o}
  """
```

Under the hood, it just calls `QueryConstruct.parse`.


shihch (approved):
LGTM

PR #48, opened 2021-02-16
add 'table unit' support

shihch:


pepegar (approved):

pepegar:
👍

JNKHunter:
If supporting the `table unit` construct doesn't sit well with us for the long term, we can revisit later.

PR #49, opened 2021-02-17
Construct support

pepegar:
This PR adds support for CONSTRUCT queries.

I've ignored one of the tests, because I need to still work on it, since the result it's returning is not what I expect.

https://github.com/gsk-aiops/bellman/compare/construct-support?expand=1#diff-00cff14d890e1d2c20ecd6793ff453d1d856242c402fa1d5e58720eadc9f7bb1R131-R186

What the engine is returning is each row duplicated.  By reading the spec, I feel that CONSTRUCT query results shouldn't have duplicated, since it explicitly talks about `set union`, not `multiset union`, and we could safely remove duplicates from result, but am not completely sure.

If you @shihch @JNKHunter know how should we proceed in that regard I'd really appreciate that.

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/QueryExecutor.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/QueryExecutor.scala
@@ -1,0 +1,43 @@
+package com.gsk.kg.engine
+
+import com.gsk.kg.sparqlparser.Query
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.SQLContext
+import com.gsk.kg.sparqlparser.Query.Construct
+
+object QueryExecutor {
+
+  def execute(
+      query: Query
+  )(df: DataFrame)(implicit sc: SQLContext): DataFrame =
+    query match {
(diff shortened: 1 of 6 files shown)


JNKHunter:
@pepegar We can safely remove duplicates. The count of final results returned should equal the cardinality of the multiset.

modules/engine/src/main/scala/com/gsk/kg/engine/QueryExecutor.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.gsk.kg.sparqlparser.Query.Construct
+
+object QueryExecutor {
+
+  def execute(
+      query: Query
+  )(df: DataFrame)(implicit sc: SQLContext): DataFrame =
+    query match {
+      case Construct(vars, bgp, r) =>
+        import sc.implicits._
```
shihch, line 15: Another option is to think the bgp in construct clause is a template, so each row in result set are transformed to new rows  through this template. I guess this way would not need to drop duplicate in the end. But not sure which way is more efficient.

shihch (approved):

PR #57, opened 2021-02-22
String functions support

pepegar:
This PR adds support for function evaluation in SparQL queries.  Most of the work is carried out by the `ExpressionF.compile` function.

# ExpressionF

The way we have the datatypes shaped up, `Expression` is a sub-language that appears inside our SparQL queries.  I've created a compiler for it there as an `Algebra` that needs a `Multiset` in order to run.

https://github.com/gsk-aiops/bellman/pull/57/files#diff-4e8bd8356385f1ba13953046603d60a9e3ceb2f42f7e5b1d42b8ece448bedfbfR126-R154


modules/engine/src/test/scala/com/gsk/kg/engine/EngineSpec.scala
```diff
@@ -198,16 +202,70 @@ class EngineSpec extends AnyFlatSpec with Matchers with DataFrameSuiteBase {
...
+        BIND(URI(CONCAT("http://lit-search-api/node/doc#", ?docid)) as ?Document) .
+      }
+      """
+
+    val inputDF = readNTtoDF("fixtures/reference-q1-input.nt")
+
+    val outputDF = readNTtoDF("fixtures/reference-q1-output.nt")
+
+    Engine.evaluate(inputDF, query) shouldBe a[Right[_, _]]
+    Engine.evaluate(inputDF, query).right.get.collect.toSet shouldEqual outputDF.collect().toSet
```
pepegar, line 250: The results we get in both sides are the same, however, we're not properly formatting result sets, as per RDF standards.

That's something I'll leave for a future PR

https://github.com/gsk-aiops/bellman/issues/58

shihch (approved):
I will re-org Expression into sealed trait at the parser side.
The angle bracket can be dropped at the parser side too, let me know which way you prefer.

PR #64, opened 2021-02-25
Re-org expressions into 3 types: StringVal, BuildInFunc and Conditional

shihch:
Gather all function expressions in BuildInFunc
StringVal is for literals
Conditional is for logical combinators of expressions


pepegar (approved):
It looks much better this way.  Thanks @shihch !

PR #66, opened 2021-02-25
DAG

pepegar:
This PR introduces an intermediate datatype to which Exprs are
compiled, `DAG`.

This datatype allows for simple rewrites using the `rewrite` method,
and centralizes all different SparQL features, while allowing for a
separation between the parser representation and the engine
representation.

## Compiler

It also introduces a `Compiler` object that will document much better
the whole structure of the project.  Looking at `Compiler.compile` one
can see much better what are all the different phases that the
compiler has, so far `parser`, `transformToGraph`, `optimizer` (a
simple `id` function for now), and `engine`.

All tests previously living in `EngineSpec` have been moved to
`CompilerSpec`.


shihch (approved):
Nicely done.
Eager to see some of implementations of the optimizer.

PR #70, opened 2021-02-26
add sparql syntax on dataframe

pepegar:
Make it possible to query a dataframe directly adding an extension method `sparql`.

``` scala
import com.gsk.kg.engine.syntax._

val df: DataFrame = ???

df.sparql("SELECT {?s ?p ?o} WHERE {?s ?p ?o}")
```


JNKHunter (approved):
LGTM

Issue #73, opened 2021-03-01
Support LIMIT in SparQL queries


pepegar:
We need to support the `LIMIT` clause in SparQL queries soon,  as they'll allow users discover datasets without loading the whole KG.

PR #74, opened 2021-03-01
added limit operation support for sparql queries on compiler

patofp:
This PR adds SparQL Solution Sequence Modifier LIMIT support.

Acceptance criteria:

If LIMIT greater than 0, will return a maximum of LIMIT elements.
If LIMIT equals 0, will return no elements.
If LIMIT less than 0, an Engine Error will happen.
If LIMIT is greater than Java MAX_INTEGER, an Engine Error will happen. (Due to Spark limit type Int)

WIP: Offset support

CLOSES: #73


pepegar (approved):

shihch (approved):
Looks good!

PR #77, opened 2021-03-02
Sparql offset support

patofp:
This PR adds SparQL Solution Sequence Modifier OFFSET support.

Acceptance criteria:

Given a sequence of solutions S = (S0, S1, ..., Sn) where N is the total number of solutions without modifier.
Then Offset(S, K) where K is the offset:
- If K >= 0 and N >= K it will return S' where S' = S(Sk, Sk+1, ..., Sn)
- If K >= 0 and N < K it will return S' where S' = ()
- If K == 0 it will return S' where S' = S
- If K < 0 it will return an Engine Error

Merge after #75

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
@@ -58,12 +58,12 @@ object ToTree extends LowPriorityToTreeInstances0 {
           case DAG.Offset(offset, r) =>
             Node(
               "Offset",
-              Stream(Leaf(offset.toString), r)
+              Stream(offset.toTree, r)
             )
           case DAG.Limit(limit, r) =>
             Node(
               "Limit",
-              Stream(Leaf(limit.toString), r)
+              Stream(limit.toTree, r)
             )
           case DAG.Distinct(r) => Node("Distinct", Stream(r))
           case DAG.Noop(str)   => Leaf(s"Noop($str)")
--- a/modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
+++ b/modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
@@ -327,6 +327,84 @@ class CompilerSpec extends AnyFlatSpec with Matchers with DataFrameSuiteBase {
     result.left.get shouldEqual EngineError.NumericTypesDoNotMatch("2147483648 to big to be converted to an Int")
   }
 
+  it should "query a real DF with offset greater than 0 and obtain a non empty set" in {
+    import sqlContext.implicits._
+
+    val df: DataFrame = List(
+      ("a", "b", "c"),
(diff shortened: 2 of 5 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
```diff
@@ -55,10 +55,15 @@ object ToTree extends LowPriorityToTreeInstances0 {
           case DAG.Filter(funcs, expr) =>
             Node("Filter", funcs.map(_.toTree).toStream #::: Stream(expr))
           case DAG.Join(l, r) => Node("Join", Stream(l, r))
-          case DAG.OffsetLimit(offset, limit, r) =>
+          case DAG.Offset(offset, r) =>
             Node(
-              "OffsetLimit",
-              Stream(Leaf(offset.toString), Leaf(limit.toString), r)
+              "Offset",
+              Stream(Leaf(offset.toString), r)
```
pepegar, line 61: I think you can change this to:

```suggestion
              Stream(offset.toTree, r)
```

If you've instances for `Show[Int]`

modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
```diff
@@ -55,10 +55,15 @@ object ToTree extends LowPriorityToTreeInstances0 {
...
+            )
+          case DAG.Limit(limit, r) =>
+            Node(
+              "Limit",
+              Stream(Leaf(limit.toString), r)
```
pepegar, line 66: Same here

```suggestion
              Stream(limit.toTree, r)
```

modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
```diff
@@ -327,6 +327,86 @@ class CompilerSpec extends AnyFlatSpec with Matchers with DataFrameSuiteBase {
...
+
+    val result = Compiler.compile(df, query)
+
+    result shouldBe a[Right[_, _]]
+    result.right.get.collect.length shouldEqual 0
+    result.right.get.collect.toSet shouldEqual Set.empty
+  }
+
+  it should "query a real DF with offset less than 0 and obtain an error result" ignore {}
+
```
pepegar, line 409: Can we delete this?
```suggestion
```

pepegar (approved):
LGTM, added a few minor comments

modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
patofp: Indeed

modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
patofp: Cool! It worked!

PR #80, opened 2021-03-03
added show typeclass to TreeRep draw method

patofp:
This PR parametrizes ToTree[A] so the nodes of the tree now can hold an A type. 
With this we can draw more generic tree representations when provided an evidence of typeclass Show on A when draw method is called.


pepegar (approved):
:clap: :clap:

PR #95, opened 2021-03-04
Reformat resulting DataFrame as per RDF

### LINKED ISSUE #58: format CONSTRUCT output as per RDF
Whenever we return another graph out of a query, we need to make sure that it's formatted correctly as per RDF standards.

pepegar:
Closes #58


patofp (approved):
LGTM!

PR #109, opened 2021-03-05
add a way of generating random DataFrames

### LINKED ISSUE #99: Come up with a way to generate test data
We need a way of generating dataframes for unit testing.

Scalacheck can be useful

pepegar:
closes #99

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,6 +15,7 @@ lazy val Versions = Map(
   "jackson"            -> "2.12.1",
   "scalacheck"         -> "1.15.2",
   "scalatestplus"      -> "3.2.3.0",
+  "sansa"              -> "0.7.1",
 )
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -32,7 +33,7 @@ inThisBuild(List(
 ))
 
 lazy val buildSettings = Seq(
-  scalaVersion := Versions("scala212"),
+  scalaVersion := Versions("scala211"),
```
pepegar, line 36: I've changed the default version of the library to 2.11.  That won't change anythying and both versions will still be published, but at least we know that we use the most restrictive one for day to day programming.  Thins such as single abstract methods or trailing commas won't be an issue anymore

patofp (approved):
GJ!

PR #111, opened 2021-03-05
modify how URIs are being matched

### LINKED ISSUE #103: After some string function calls, either missing '<' at front or extra '>' at the end
For example:
URI(CONCAT("http://lit-search-api/node/doc#", ?docid) 
missed '<' in the result set
while
STRAFTER(str(?d), "#")
has extra '>' at the end since ?d is an URI

pepegar:
use the the _Jena Standard_, and unquote URIs before matching.  This
way we'll avoid having errors in which the angle brackets are
understood as part of the URI.

## Change the way initial dataframes are being loaded.

This PR changes the way our engine is matching URIs, since preiously it was matching the wrapping angle brackets, while now it's not.  Now we should either

1. remove angle brackets manually from all fields (this can be a bit annoying but something like `field.stripPrefix("<").stripSuffix(">")` should work.
2. Load the DataFrame from the NT file with either Jena or sansa-rdf.

```
// Example with Jena

  private def readNTtoDF(path: String) = {
    import sqlContext.implicits._
    import scala.collection.JavaConverters._

    val filename = s"modules/engine/src/test/resources/$path"
    val inputStream: CollectorStreamTriples = new CollectorStreamTriples();
    RDFParser.source(filename).parse(inputStream);

    inputStream
      .getCollected()
      .asScala
      .toList
      .map(triple =>
        (triple.getSubject().toString(), triple.getPredicate().toString(), triple.getObject().toString())
      ).toDF("s", "p", "o")

  }
```

Closes #103


patofp (approved):
I hope you don't have nightmares this weekend with angle brackets :P

PR #114, opened 2021-03-08
added sparql filter expression support

### LINKED ISSUE #83: Add support for FILTER
We should add support for queries that filter their results

An example here
https://github.com/gsk-aiops/bellman/blob/6327df4d52c367ae6075200628a3532c4bb0ab02/modules/parser/src/test/scala/com/gsk/kg/sparqlparser/QuerySamples.scala#L55

patofp:
This PR add SparQL FILTER support.

It restricts the solution of a graph according to a given constraint. This is done by eliminate any solutions that, when evaluated  into the expression, either result in an effective boolean value of false or produce an error.

Closes #83

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
@@ -1,18 +1,16 @@
 package com.gsk.kg.engine
 
 import cats.Foldable
+import cats.data.NonEmptyList
 import cats.instances.all._
+import cats.syntax.EitherSyntax
 import cats.syntax.either._
 import cats.syntax.applicative._
-
-import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.{Column, DataFrame, SQLContext}
 import org.apache.spark.sql.functions._
-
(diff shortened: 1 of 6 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -60,6 +56,19 @@ object Engine {
       .map(_.dataframe)
   }
 
+  private def evaluateFilter(funcs: List[Expression], expr: Multiset): M[Multiset] = {
+
+    val compiledFuncs: List[DataFrame => Result[Column]] = funcs.map(ExpressionF.compile[Expression])
+
+    M.liftF[Result, DataFrame, Multiset] {
+      // TODO: Do it for all the compiled Funcs
+      compiledFuncs.headOption.toRight(EngineError.FunctionError("No functions on filter"))
```
pepegar, line 65: Seeing this error makes me think... Shouldn't we change the type of `funcs` from `List[Expression]` to `NonEmptyList[Expression]`? We'd add type safety and avoid juggling optionals and errors here.

pepegar (changes requested):
I just added that small comment.  what do you think @patofp , should we change it?

PR #117, opened 2021-03-09
added sparql optional support

### LINKED ISSUE #87: Add support for OPTIONAL
Including multiple optionals

SELECT ?s ?s_name ?o ?o_name WHERE {
        ?s <http://www.w3.org/2000/01/rdf-schema#subClassOf> ?o . 
        FILTER (!isBlank(?s) && !isBlank(?o)) .
        OPTIONAL {?s <http://www.w3.org/2000/01/rdf-schema#label> ?s_name}
        OPTIONAL {?o <http://www.w3.org/2000/01/rdf-schema#label> ?o_name}
    }

patofp:
This PR adds SparQL support for OPTIONAL operation. It performs a left outer join in Spark and also applies, in case they exist, filters on the right side of the join.

Closes #87

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
@@ -34,7 +34,7 @@ object Engine {
         evaluateBind(variable, expression, r)
       case DAG.Triple(s, p, o)         => evaluateTriple(s, p, o)
       case DAG.BGP(triples)            => Foldable[List].fold(triples).pure[M]
-      case DAG.LeftJoin(l, r, filters) => notImplemented("LeftJoin")
+      case DAG.LeftJoin(l, r, filters) => evaluateLeftJoin(l, r, filters)
       case DAG.Union(l, r)             => l.union(r).pure[M]
       case DAG.Filter(funcs, expr)     => evaluateFilter(funcs, expr)
       case DAG.Join(l, r)              => notImplemented("Join")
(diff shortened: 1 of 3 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -58,6 +58,14 @@ object Engine {
       .map(_.dataframe)
   }
 
+  private def evaluateLeftJoin(l: Multiset, r: Multiset, filters: List[Expression]): M[Multiset] = {
+    if (filters.isEmpty) {
+      M.liftF[Result, DataFrame, Multiset](l.leftJoin(r))
+    } else {
+      evaluateFilter(NonEmptyList.fromListUnsafe(filters), r).flatMapF(l.leftJoin)
```
pepegar, line 65: I'd try to avoid all `*Unsafe` operators as much as possible in our code.  You could use `fromList`, which returns an option and then lift that to `M`?

pepegar (changes requested):
Just a small change.  Once that's fixed, LGTM!

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
patofp: Yeah, I was able to lift it. Thanks @pepegar! Much better now

pepegar (approved):
good job  mate!

PR #129, opened 2021-03-12
make blank nodes in templates work

### LINKED ISSUE #107: Blank node in construct should produce unique blank node label per solution
As per spec:
https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#templatesWithBNodes

_"A template can create an RDF graph containing blank nodes. The blank node labels are scoped to the template for each solution. If the same label occurs twice in a template, then there will be one blank node created for each query solution, but there will be different blank nodes for triples generated by different query solutions."_

Currently, is seems our solution is creating a single blank node (using a UUID), and reusing for each solution in the result.
image

pepegar:
This PR changes the way we're evaluating CONSTRUCT templates.  Now we're iterating over all solutions in the multiset and, for each one of them, we're creating as many triples in the final result as described in the template.

This helps with creating a different ID per solution for blank nodes.

Closes #107


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -84,45 +95,71 @@ object Engine {
...
+            .flatMap(extractBlanks)
+            .distinct
+            .map(blankLabel => (blankLabel.s, ju.UUID.randomUUID().toString()))
+            .toMap
+
+        templateValues.map { triple =>
+          val fields: List[Any] = triple
+            .map({
+              case (VARIABLE(s), pos) =>
+                (solution.get(solution.fieldIndex(s)), pos)
```
patofp, line 146: I understand this .get over the Row with the solution index will be always safe because that index will exists. Is that correct?

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -84,45 +95,71 @@ object Engine {
...
+              case (BLANK(x), pos) =>
+                (blankNodes.get(x).get, pos)
```
patofp, line 148: This .get is over a Map and can be unsafe, is it posible to use .getOrElse instead? I understand that the `x` will act as the key but not sure if it safe in this context. Also if not possible the value could be extracted simply with `blankNodes(x)`

patofp (approved):
I left you two comments for review, for the rest. GJ @pepegar !

PR #139, opened 2021-03-16
Compact bgp queries

pepegar:


modules/engine/src/main/scala/com/gsk/kg/engine/data/ChunkedList.scala
```diff
@@ -0,0 +1,111 @@
...
+  * can be [[ChunkedList.Chunk]]s of elements.
+  */
+@deriveTraverse trait ChunkedList[A] {
+
+  final def compact[B](f: A => B)(implicit B: Order[B]): ChunkedList[A] =
+    ChunkedList.fromChunks(groupBy(f).values.toList)
+
+  def mapChunks[B](fn: Chunk[A] => B): ChunkedList[B] =
+    this match {
+	    case Empty() => Empty()
```
pepegar, line 25: format this!

modules/engine/src/main/scala/com/gsk/kg/engine/data/ChunkedList.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  final def reverse: ChunkedList[A] =
+    this.foldLeft[ChunkedList[A]](Empty()) { (acc, elem) =>
+      NonEmpty(Chunk(elem), acc)
+    }
+
+  final def concat(other:ChunkedList[A]): ChunkedList[A] =
+    this.reverse.foldLeftChunks[ChunkedList[A]](other) { (acc, elem) =>
+      acc match {
+	      case Empty() => Empty()
```
pepegar, line 64: Also this

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -66,6 +68,38 @@ object Engine {
...
+    M.get[Result, DataFrame].map { df =>
+      Foldable[ChunkedList].fold(
+        triples.mapChunks { chunk =>
+          val condition: Column =
+            chunk
+              .map(_.getPredicates)
+              .map(
+                _.map({ case (pred, position) =>
+                  df(position) === pred.s
+                }).foldLeft(lit(true))((acc, current) => acc && current)
```
patofp, line 83: I think this `foldLeft` could be replaced by `.forall(_ == true)`. The advantage of forall is that it will cut iteration once it finds first `false`.

modules/engine/src/main/scala/com/gsk/kg/engine/data/ChunkedList.scala
```diff
@@ -0,0 +1,111 @@
...
+import cats.data.NonEmptyChain
+import java.{util => ju}
+import scala.collection.immutable.SortedMap
+import scala.collection.immutable.Nil
+
+/**
+  * A data structure like a linked [[scala.List]] but in which nodes
+  * can be [[ChunkedList.Chunk]]s of elements.
+  */
+@deriveTraverse trait ChunkedList[A] {
```
patofp, line 18: Awesome @pepegar!

patofp (approved):
GJ @pepegar!

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
pepegar: I don't think so, this `&&` is `org.apache.spark.sql.Column`'s, we need to merge all the queries into a single condition with `AND`.

patofp: Oh, ok then ;)

PR #142, opened 2021-03-17
format localized strings correctly

### LINKED ISSUE #134: Unexpected output for string literals. Extra outer quotes.
Quotes for this query output seems incorrect:

image

Should be:
"NICR2579"@en
"NICR2579 \u7EC6\u80DE"@zh

**Query**
```
SELECT ?label WHERE {
        <http://purl.obolibrary.org/obo/CLO_0052588>
        <http://www.w3.org/2000/01/rdf-schema#label>
        ?label.
    }
```

I've placed the input file ont.nt on DB cluster for testing

pepegar:
Identify correctly localized strings so that they get formatted without extra quotes surrounding them.

Closes #134


patofp (approved):
LGTM!

PR #143, opened 2021-03-17
introduce RemoveNestedProject optimization

### LINKED ISSUE #138: Remove nested Project in DAG
Currently, when parsing a `SELECT` query, we produce two nested Project nodes in the DAG.

This is because the `Expr` itself produce a `Project`, and the `SELECT` part of the query another.

We can create a rewrite rule for flattening nested `Project`s.

pepegar:
This optimization will deduplicate Project nodes in the DAG

Closes #138


patofp (approved):
LGTM

Issue #132, opened 2021-03-12
Logical error handling: Fail quickly when SELECT/CONSTRUCT vars do not exist in WHERE clause


JNKHunter:
Expected behavior: Query should fail and exit before execution begins if variable is referenced in SELECT or CONSTRUCT that does not exist in WHERE clauses.

Current behavior: Spark execution begins and only fails when execution of WHERE is complete.

Example 
SELECT ?species_node WHERE {
        <http://purl.obolibrary.org/obo/CLO_0037232> <http://www.w3.org/2000/01/rdf-schema#subClassOf> ?derived_node .
}

![image](https://user-images.githubusercontent.com/2665766/111003327-e8228500-8354-11eb-8b93-8c8f97510607.png)

PR #144, opened 2021-03-17
introduce optics

pepegar:
This commit introduces monocle-core and monocle-macro libraries, that
will allow us for easier manipulation of datatypes.

## Law testing

This PR also introduces Law testing, that we're implementing for monocle optics, but we could implement for all other datatypes & typeclasses that support it.

Laws are sets of tests create by library authors as Scala code, such as [this PrismTest by Monocle](https://github.com/optics-dev/Monocle/blob/master/law/src/main/scala/monocle/law/discipline/PrismTests.scala) that everyone can run as part of their codel.  In our case, we're testing all our optics given the specification that Monocle authors provided.

Will help us with #132 and other static analysis related issues.


patofp (approved):
Optics are going to be very useful to do optimizations. Great idea @pepegar! LGTM

PR #158, opened 2021-03-19
introduce analyzer phase

### LINKED ISSUE #132: Logical error handling: Fail quickly when SELECT/CONSTRUCT vars do not exist in WHERE clause

pepegar:
In this phase we'll catch logical errors such as the one described in #132.

Closes #132


modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/Analyzer.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+    * Execute all rules in [[Analyzer.rules]] and accumulate errors
+    * that they may throw.
+    *
+    * In case no errors are returned, the
+    *
+    * @return
+    */
+  def analyze[T: Basis[DAG, *]]: Phase[T, T] =
```
patofp, line 32: Great idea using Validated this way we can stack validation errors at once!

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/package.scala
```diff
@@ -0,0 +1,9 @@
+package com.gsk.kg.engine
+
+import cats.data.ValidatedNec
+
+package object analyzer {
+
+  type Rule[T] = T => ValidatedNec[String, String]
```
patofp, line 7: Nice!

patofp (approved):
LGTM!

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/Analyzer.scala
```diff
@@ -0,0 +1,68 @@
...
+
+      x match {
+        case Invalid(e) => M.lift[Result, DataFrame, T](EngineError.AnalyzerError(e).asLeft)
+        case Valid(e) => t.pure[M]
+      }
+    }
+
+  def findUnboundVariables[T](implicit T: Basis[DAG, T]): Rule[T] = { t =>
+    val declaredVariables: Set[VARIABLE] = t.collect[List[List[VARIABLE]], List[VARIABLE]] {
+        case BGP(triples) =>
```
shihch, line 44: Besides BGP, Extend, FilterLeftJoin, Filter, and nested Project could introduce variables. Do we have to consider all of them?

pepegar: Thanks @shihch good point.

I've changed the approach for doing the analysis to a full bottom-up traverse checking variables declared level by level

patofp (approved):
LGTM!

PR #168, opened 2021-03-25
sparql support for graph joins

patofp:
This PR add support for joining two graphs. Example:

```

SELECT ?x ?mbox
WHERE
{
   GRAPH ex:alice { ?x foaf:mbox ?mbox . }
   GRAPH ex:bob { ?x foaf:mbox ?mbox . }
}
```

It also adds some missing test on `Multiset` operations.

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
@@ -7,8 +7,11 @@ import cats.syntax.all._
 import org.apache.spark.sql.Column
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.functions._
 
+import com.gsk.kg.engine.Multiset.filterGraph
+import com.gsk.kg.sparqlparser.StringVal.GRAPH_VARIABLE
 import com.gsk.kg.sparqlparser.StringVal.VARIABLE
 
 /** A [[Multiset]], as expressed in SparQL terms.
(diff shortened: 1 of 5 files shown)


pepegar (approved):
LGTM, great job!

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -37,40 +42,85 @@ final case class Multiset(
...
-          .join(b.dataframe.as("b"), common.map(_.s).toSeq)
-        Multiset(a.bindings.union(b.bindings), df)
-    }
+  def join(other: Multiset): Multiset = (this, other) match {
+    case (l, r) if l.isEmpty => r
+    case (l, r) if r.isEmpty => l
+    case (l, r)
+        if (filterGraph(l).bindings intersect filterGraph(
+          r
+        ).bindings).isEmpty =>
```
pepegar, line 51: Can we extract the guard to a named private function? something like
```suggestion
        if noCommonBindingsBetween(l, r) =>
```

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -37,40 +42,85 @@ final case class Multiset(
...
+      .drop(leftGraphCol, rightGraphCol)
+      .withColumn(GRAPH_VARIABLE.s, lit(""))
+      .schema
+
+    // For each element on the array of *g column it generates a new row with the graph (unfolds *g column)
+    val result = productWithMergedGraphColunns.flatMap { r =>
+      val index  = r.fieldIndex(GRAPH_VARIABLE.s)
+      val graphs = r.getSeq[String](index)
+      graphs.distinct.map(g => Row.fromSeq(r.toSeq.dropRight(1) :+ g))
+    }(RowEncoder(resultSchema))
```
pepegar, line 96: :+1:! I'd also add a commment here noting that we're traversing the whole dataset here, and not apply a column function, so there may be a performance problem

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -209,11 +257,48 @@ final case class Multiset(
...
+  private val addDefaultGraph: Multiset => Multiset = { m =>
+    if (m.isEmpty) {
+      m
+    } else {
+      m.copy(
+        bindings = m.bindings + VARIABLE(GRAPH_VARIABLE.s),
+        dataframe = m.dataframe.withColumn(GRAPH_VARIABLE.s, lit(""))
+      )
+    }
+  }
```
pepegar, line 284: Can you add Scaladoc to these two to have them documented?

modules/engine/src/test/scala/com/gsk/kg/engine/utils/MultisetMatchers.scala
```diff
@@ -0,0 +1,39 @@
+package com.gsk.kg.engine.utils
+
+import com.gsk.kg.engine.Multiset
+
+import org.scalatest.matchers.MatchResult
+import org.scalatest.matchers.Matcher
+import org.scalatest.matchers.should.Matchers
+
+trait MultisetMatchers extends Matchers {
```
pepegar, line 9: Great idea!

pepegar (approved):
Added a couple of minor comments, looks good otherwise!

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
patofp: Totally agree! This entire method should be revisited at some time

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
patofp: Yeah, good idea

PR #169, opened 2021-03-25
make it possible to apply binary operators on dates

### LINKED ISSUE #136: Add support on logical operators for dateTime type
Based on SparQL Spec: https://www.w3.org/TR/rdf-sparql-query/#OperatorMapping, it is possible to compare DateTime type.

So we need to implement this for next operations: `=`, `!=`, `>`, `<`, `>=`, `<=`.

 Example: 

```
PREFIX : <http://example.org>

SELECT *
WHERE
{
  ?s :date ?date .
  FILTER (?date > "2014-05-23T10:20:13+05:30"^^xsd:dateTime)
}
```

pepegar:
This PR introduces the ability to use binary operators on xsd:dateTime values.

~I still need to work through some binary dependency issues with Spark in the CompilerSpec before turning it into a _real_ PR :)~

Closes #136

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -176,7 +176,7 @@ lazy val `bellman-spark-engine` = project
 
 addCommandAlias(
   "ci-test",
-  ";scalafix --check ;scalafmtCheckAll ;scalastyle ;+test"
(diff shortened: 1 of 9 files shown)


patofp (approved):
LGTM!

PR #177, opened 2021-03-29
fix-compact-bgp-ordering-issue

pepegar:
Fix triple ordering issue caused by `CompactBGPs` step


shihch (approved):
Looks good. Thanks for putting up the fix quickly.

patofp (approved):
GJ @pepegar!

PR #178, opened 2021-03-30
bring back support for SPO DataFrames

### LINKED ISSUE #167: Investigate support for SPO dataframes
Let's see how hard it is to keep supporting them, we can consider that if they don't have a `*g` column, all triples belong to the default graph.

pepegar:
Now we're performing a validation before running the Spark job so that
we can check that received DF is either 3 or 4 cols wide, otherwise we
raise an error.

IN case of a 3 col DF, we manually add the `g` one.

Closes #167


patofp (approved):
LGTM!

Issue #180, opened 2021-03-31
Rename g column of Quads inside Scan


patofp:
We need to differentiate between Quads  on the DAG that are inside a GRAPH statement. Eg:

```
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
        PREFIX dc: <http://purl.org/dc/elements/1.1/>
        PREFIX ex: <http://example.org/>
        
        SELECT ?mbox ?name
        FROM <http://example.org/dft.ttl>
        FROM NAMED <http://example.org/alice>
        FROM NAMED <http://example.org/bob>
        WHERE
        {
           GRAPH ex:alice { 
             ?x foaf:mbox ?mbox .
             ?x foaf:name ?name .
           }
        }
```

Quads inside GRAPH should hold the graph name instead of the *g variable.

PR #183, opened 2021-03-31
Named graph pushdown

pepegar:
Part of #180


patofp (approved):
Thanks @pepegar, this technique is so cool. Now we know how to pass a context through the tree in a top-down way. Adding that function to the algebra F(A -> B) -> (A -> B).

PR #194, opened 2021-04-05
implement SAMPLE aggregate Function

### LINKED ISSUE #155: Add support for SAMPLE
https://www.w3.org/TR/sparql11-query/#defn_aggSample

pepegar:
This PR implements the SAMPLE aggregate function and introduces other
aggregate functions to the AST.

Closes #155


modules/engine/src/main/scala/com/gsk/kg/engine/ExpressionF.scala
```diff
@@ -131,42 +149,46 @@ object ExpressionF {
...
       }
 
     val eval = scheme.cataM[M, ExpressionF, T, Column](algebraM)
 
     eval(t).runA(df)
   }
 
+  private def unknownFunction(name: String): M[Column] =
+    M.liftF[Result, DataFrame, Column](
+      EngineError.UnknownFunction("GROUP_CONCAT").asLeft[Column]
```
patofp, line 191: I think the message should be something meaningful for every function not implemented yet.

patofp (approved):
LGTM!

PR #196, opened 2021-04-06
Sparql default named graphs mixing

### LINKED ISSUE #171: Support queries that mixes default and named graphs
Examples:

```
SELECT ?x ?mbox
WHERE
{
   ?x foaf:mbox ?mbox .
   GRAPH ex:bob { ?x foaf:mbox ?mbox . }
}
```

```
SELECT ?x ?mbox
WHERE
{
   { ?x foaf:mbox ?mbox . }
   UNION
   { GRAPH ex:bob { ?x foaf:mbox ?mbox . } }
}
```

### LINKED ISSUE #184: Add support on Quads to contain multiple graphs
When we have a query like this:

```

SELECT ?x ?mbox ?name
FROM <http://example.org/alice>
FROM <http://example.org/charles>
WHERE
{
   { ?x foaf:name ?name }
   UNION
   { GRAPH ex:bob { ?x foaf:mbox ?mbox } }
}
```

The Quad to evaluate the first BGP should contain multiple graphs so it must contain a lis of graphs.

patofp:
This PR adds support to queries that mixes *default* and *named* graphs. Eg:

```

SELECT ?x ?y ?mbox ?name
FROM <http://example.org/alice>
FROM <http://example.org/charles>
WHERE
{
   { ?x foaf:name ?name . }
   UNION 
  { GRAPH ex:bob { ?y foaf:mbox ?mbox } }
}
```

So in the example the evaluation of first BGP Quad will contain the results from the *default* graphs ("", "http://example.org/alice", "http://example.org/charles") and the evaluation of the second BGP will contain the results of the graph "http://example.org/bob". At last the UNION operation will be applied.

As we can see this is evaluated in a bottom-up way, but the Quads of the BGPs need the some context of the graph/s they are querying so this context should be passed in a top-bottom way on the construction of the graph. This is the reason a PushDown of this info has been implemented.

Principal changes are:
- Added List of graphs to the Quads.
- Added PushDownGraph on Optimizer.
- Modified `evaluateBGP` to support Quads with a list of graphs.

Closes #171 
Closes #184


modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -46,11 +46,19 @@ final case class Multiset(
   def join(other: Multiset): Multiset = (this, other) match {
     case (l, r) if l.isEmpty => r
     case (l, r) if r.isEmpty => l
-    case (l, r) if noCommonBindings(l, r) =>
+    case (l, r)
+        if noCommonBindings(l, r) &&
+          l.dataframe.columns.contains(GRAPH_VARIABLE.s) &&
+          r.dataframe.columns.contains(GRAPH_VARIABLE.s) =>
```
pepegar, line 52: Can you extract this to a more meaningful method?

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
```diff
@@ -0,0 +1,44 @@
...
+import higherkindness.droste.Basis
+import higherkindness.droste.scheme
+
+import com.gsk.kg.engine.DAG
+import com.gsk.kg.sparqlparser.StringVal
+import com.gsk.kg.sparqlparser.StringVal.URIVAL
+
+object GraphsPushdown {
+
+  def apply[T](implicit T: Basis[DAG, T]): (T, List[StringVal]) => T = {
```
pepegar, line 13: Can you add scaladoc explaining what does this optimization do?

modules/parser/src/main/scala/com/gsk/kg/sparql/syntax/Interpolators.scala
```diff
@@ -2,12 +2,13 @@ package com.gsk.kg.sparql.syntax
...
 import com.gsk.kg.sparqlparser.Query
 import com.gsk.kg.sparqlparser.QueryConstruct
+import com.gsk.kg.sparqlparser.StringVal
 
 trait Interpolators {
 
   implicit class SparqlQueryInterpolator(sc: StringContext) {
 
-    def sparql(args: Any*): Query = {
+    def sparql(args: Any*): (Query, List[StringVal]) = {
```
pepegar, line 11: I'd prefer keeping this returing only the `Query`.  We can still access the list of graphs from the `QueryConstruct` object.

pepegar (approved):
Looks good to me, just added a couple of minor comments!

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -110,24 +112,32 @@ object Engine {
   )(implicit sc: SQLContext): M[Multiset] = {
     import sc.implicits._
     import org.apache.spark.sql.functions._
+
+    def composedConditionFromChunk(
```
pepegar, line 116: Can you also add a small explaination to what we're doing here? I think it's becoming more and more arcane hahaha

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
patofp: Done ;)

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
patofp: Indeed

modules/parser/src/main/scala/com/gsk/kg/sparql/syntax/Interpolators.scala
patofp: I found it useful for testing purposes because when parsed a query this way we can extract also the default graphs and use them to check some conditions on testing without having to call twice to the QueryConstruct.parse method. Eg:

```
val (query, defaultGraphs) =
          sparql"""
    
            SELECT ?mbox ?name
            WHERE
            {
               GRAPH ex:alice {
                 ?x foaf:mbox ?mbox .
                 ?x foaf:name ?name .
               }
            }
          """

defaultGraphs shouldEqual List(URIVAL(""), URIVAL("http://example.org/dft.ttl"))
```

I think we can leave the *sparql* to return only the query and create another Interpolator *sparqlWithGraphs* that returns both. What do you think?

pepegar: In the end, if we want all the data, we can use `QueryConstruct.parse(whatever)` (that's what `sparql"""xxx"""` is using under the hood).

I'd rather not introduce more interpolators unless we actually find a stronger case for them, but that's just my opinion.

patofp: Yeah, no need to add that Interpolator, only one call to QueryConstruct.parse is needed. ;)

PR #199, opened 2021-04-08
Implement GROUP BY and some aggregate Functions

### LINKED ISSUE #151: Add support for GROUP BY, and COUNT
https://www.w3.org/TR/sparql11-query/#groupby
https://www.w3.org/TR/sparql11-query/#defn_aggCount

Has to be completed after #155, given the evaluation semantics of GROUP BY

### LINKED ISSUE #153: Add support for MIN and MAX
https://www.w3.org/TR/sparql11-query/#defn_aggMin
https://www.w3.org/TR/sparql11-query/#defn_aggMax

### LINKED ISSUE #154: Add support for AVG and SUM
https://www.w3.org/TR/sparql11-query/#defn_aggAvg
https://www.w3.org/TR/sparql11-query/#defn_aggSum

pepegar:
This PR introduce GROUP BY statements to belman, alongside some aggregation functions:

- `max` #153
- `min` #153
- `avg` #154
- `sum` #154
- `count` #151

Closes #153
Closes #154
Closes #151


modules/engine/src/main/scala/com/gsk/kg/engine/Func.scala
```diff
@@ -205,6 +205,9 @@ object Func {
   def sample(col: Column): Column =
     first(col, true)
 
+  def groupConcat(col: Column, separator: String): Column =
+    ???
```
patofp, line 209: Could this generate an exception?

modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
```diff
@@ -3874,6 +3874,256 @@ class CompilerSpec extends AnyWordSpec with Matchers with DataFrameSuiteBase {
...
+            ?a ?b <http://uri.com/object>
+          } GROUP BY ?a
+          """
+
+        val result = Compiler.compile(df, query)
+
+        result.right.get.collect should have length 3
+      }
+
+      "operate correctly there's GROUP BY and a HAVING clause" ignore {}
```
patofp, line 4125: We can set a TODO with the issue that implements the HAVING clause. What do you think?

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/QueryConstruct.scala
```diff
@@ -56,7 +56,9 @@ object QueryConstruct {
   }
 
   private def getVars(query: org.apache.jena.query.Query): Seq[VARIABLE] =
-    query.getProjectVars.asScala.map(v => VARIABLE(v.toString()))
+    query.getProjectVars.asScala.map(v =>
+      VARIABLE(v.toString().replace(".", ""))
```
patofp, line 60: Just curious. Which is the reason of replacing dots? 🤔

patofp (approved):
GJ @pepegar! Just a couple of comments 😉

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/QueryConstruct.scala
pepegar: Oh, yeah!  Spark doesn't allow dots in column names because of spark SQL.  I run this _replace_ because Jena generates temporary variable names like this: `$.0`, `$.1`...

modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
pepegar: I have removed it altogether, we gain nothing by having just the test name there.  Good catch ;)

modules/engine/src/main/scala/com/gsk/kg/engine/Func.scala
pepegar: Currently we don't have queries that execute `GROUP_CONCAT`, but it could indeed.   I'll create an issue for that

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/QueryConstruct.scala
patofp: oh, good to know ;)

PR #206, opened 2021-04-12
added support for variables on graph statement

### LINKED ISSUE #173: Support variables on GRAPH statement
Example:

```

SELECT ?x ?mbox ?g
WHERE
{
   GRAPH ?g { ?x foaf:mbox ?mbox }
}
```

patofp:
This PR adds support for use of variables in GRAPH statement. Eg:

```

SELECT ?who ?g ?mbox
FROM NAMED <http://example.org/charles>
WHERE
{
   ?g dc:publisher ?who .
   GRAPH ?g { ?x foaf:mbox ?mbox }
}
```

Closes #173

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
@@ -103,11 +103,14 @@ object Engine {
     l.union(r).pure[M]
 
   private def evaluateScan(graph: String, expr: Multiset): M[Multiset] = {
-    val df = expr.dataframe
-      .filter(expr.dataframe(GRAPH_VARIABLE.s) === graph)
-      .withColumn(GRAPH_VARIABLE.s, lit(""))
-    expr.copy(dataframe = df).pure[M]
-  }
+    val bindings =
+      expr.bindings.filter(_.s != GRAPH_VARIABLE.s) + VARIABLE(graph)
+    val df = expr.dataframe.withColumn(graph, expr.dataframe(GRAPH_VARIABLE.s))
+    Multiset(
+      bindings,
+      df
+    )
+  }.pure[M]
 
   private def evaluateBGP(
       quads: ChunkedList[Expr.Quad]
@@ -119,7 +122,11 @@ object Engine {
           val condition = composedConditionFromChunk(df, chunk)
           val current   = df.filter(condition)
           val vars =
-            chunk.map(_.getNamesAndPositions).toChain.toList.flatten
+            chunk
+              .map(_.getNamesAndPositions :+ (GRAPH_VARIABLE, "g"))
+              .toChain
+              .toList
+              .flatten
           val selected =
             current.select(vars.map(v => $"${v._2}".as(v._1.s)): _*)
 
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
@@ -96,7 +96,9 @@ object Engine {
     )
   )
 
-  private def evaluateJoin(l: Multiset, r: Multiset): M[Multiset] =
+  private def evaluateJoin(l: Multiset, r: Multiset)(implicit
+      sc: SQLContext
+  ): M[Multiset] =
     l.join(r).pure[M]
 
   private def evaluateUnion(l: Multiset, r: Multiset): M[Multiset] =
@@ -116,6 +118,7 @@ object Engine {
       quads: ChunkedList[Expr.Quad]
   )(implicit sc: SQLContext): M[Multiset] = {
     import sc.implicits._
+
     M.get[Result, DataFrame].map { df =>
       Foldable[ChunkedList].fold(
         quads.mapChunks { chunk =>
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
@@ -8,7 +8,6 @@ import org.apache.spark.sql.Column
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.SQLContext
-import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.functions._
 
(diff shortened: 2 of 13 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -119,7 +122,11 @@ object Engine {
           val condition = composedConditionFromChunk(df, chunk)
           val current   = df.filter(condition)
           val vars =
-            chunk.map(_.getNamesAndPositions).toChain.toList.flatten
+            chunk
+              .map(_.getNamesAndPositions :+ (GRAPH_VARIABLE, "g"))
```
pepegar, line 126: Why are we appending the GRAPH_VARIABLE here?

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -370,6 +362,88 @@ object Multiset {
...
+    val innerJoin = renamedLeft.join(renamedRight, intersect, "inner")
+
+    // Merges the *l and *r columns into a *g column with Array(l, r)
+    val innerWithMergedGraphColunns = innerJoin
+      .withColumn(GRAPH_VARIABLE.s, array(leftGraphCol, rightGraphCol))
+      .drop(leftGraphCol, rightGraphCol)
+
+    // Generates a schema for the final DF (needed for the flatMap)
+    val resultSchema = innerWithMergedGraphColunns
+      .drop(leftGraphCol, rightGraphCol)
```
pepegar, line 422: Do we need to drop the columns twice?
```suggestion

    // Generates a schema for the final DF (needed for the flatMap)
    val resultSchema = innerWithMergedGraphColunns
      .drop(leftGraphCol, rightGraphCol)
```

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -370,6 +362,88 @@ object Multiset {
...
+      }
+    }(RowEncoder(resultSchema))
+
+    // Needed to keep Schema information on each Row. We will have GenericRowWithSchema instead of GenericRow
+    SparkSession
+      .builder()
+      .getOrCreate()
+      .sqlContext
+      .sparkSession
+      .createDataFrame(result.toJavaRDD, resultSchema)
```
pepegar, line 444: I'd avoid calling getting the sqlContext this way (in the whole file actually).  Can we just pass it as a parameter, implicit maybe?

We have it available in the Engine already, we wouldn't even need to change the call-site

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
```diff
@@ -52,7 +52,9 @@ object FindUnboundVariables {
           declared <- State.get
         } yield (used diff declared) ++ r
       case Scan(graph, expr) =>
-        Set.empty.pure[ST]
+        State
+          .modify[Set[VARIABLE]](x => x + VARIABLE(graph))
```
pepegar, line 56: 👌🏻 good catch!

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
```diff
@@ -113,37 +116,56 @@ import com.gsk.kg.sparqlparser.StringVal.URIVAL
...
+          case DAG.Describe(vars, r) =>
+            graphsOrList => DAG.describeR(vars, r(graphsOrList))
+          case DAG.Ask(r) => graphsOrList => DAG.askR(r(graphsOrList))
+          case DAG.Construct(bgp, r) =>
+            graphsOrList => DAG.constructR(bgp, r(graphsOrList))
           case DAG.Scan(graph, expr) =>
-            graphs => expr(URIVAL(graph) :: Nil)
+            _ =>
+              if (graph.startsWith("?")) {
+                DAG.scanR(graph, expr(graphs.named.asLeft))
```
pepegar, line 133: Don't we have an option of deleting this node from the graph?

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
```diff
@@ -113,37 +116,56 @@ import com.gsk.kg.sparqlparser.StringVal.URIVAL
...
 object GraphsPushdown {
 
-  def apply[T](implicit T: Basis[DAG, T]): (T, List[StringVal]) => T = {
-    case (t, defaultGraphs) =>
-      val alg: Algebra[DAG, List[StringVal] => T] =
-        Algebra[DAG, List[StringVal] => T] {
-          case DAG.Describe(vars, r) => graphs => DAG.describeR(vars, r(graphs))
-          case DAG.Ask(r)            => graphs => DAG.askR(r(graphs))
-          case DAG.Construct(bgp, r) => graphs => DAG.constructR(bgp, r(graphs))
+  type GraphsOrList = Either[List[StringVal], Graphs]
```
pepegar, line 119: What does this represent? I mean, if `Graphs` already contain `defaults` and `named`, what's the idea behind this either?

pepegar (commented):
Added a couple of comments, but this looks awesome!  Great job buddy

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
patofp: This is needed to keep track of the graph column when performing later operations such as JOIN or UNION when mixing default and named graphs BGPs. Eg:

```
?g dc:publisher ?who . // <- BGP-1
 GRAPH ?g { ?x foaf:mbox ?mbox } // BGP-2
```

If we don't have the `*g` column there we will loose the graph info of the BGP-1

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
patofp: In this case we are generating an schema to use later in the .flatMap as RowEncoder, so we construct the schema based on the input Dataset, but the output Dataset from flatmap doesn't have that columns. Do you think there is a better way of generating the output schema?

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
patofp: Cool, I'll change this!

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
patofp: Right now I don't see anyway we can get rid of this node when we have a GRAPH statement with variable. I think we should revisit this at some time though.

So the explanation is that imagine we have the next WHERE clause:

```
?g dc:publisher ?who .
GRAPH ?g { ?x foaf:mbox ?mbox }
```

This will generate a DAG -> `Join(BGP(defaultQuads), Scan(BGP(namedQuads)))`, the Dataframe that is generated when evaluated BGP(namedQuads) will contain columns (?x, ?mbox, *g) so when Scan is evaluated it will rename the `*g` column to the variable binding `?g`.

So `Scan` will replace the `*g` to the variable that the user used in the query, if not that info of the variable is lost.

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
patofp: So when pushingdown the `Graphs(default: List[StringVal], named: List[StringVal])` at the point where we reach a BGP node we must decide whether we use `default` or `named` graphs. This disjunction is abstracted by the Either type in the way that if it is a Left, then we know that Quad is inside a GRAPH statement so it must take the named graphs, if it is Right then we use the default graphs at the Right projection. 

Does it make sense? Maybe there is a better way to abstract this.

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
pepegar: Oh no, I just meant that we had already dropped leftgraphcol and rightgraphcol from innerWithMergedGraphColunns

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
pepegar: We can leave this as it is, and change that later on

pepegar (approved):

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
patofp: oohhh I see, good catch mate!

modules/engine/src/main/scala/com/gsk/kg/engine/optimize/GraphsPushdown.scala
patofp: Sure

PR #224, opened 2021-04-16
added strip question marks on ouput

### LINKED ISSUE #195: Strip question marks from column names
By default, strip question marks from column names returned from a where block. Give the user the ability to include the question marks with a flag that is passed during the method call: preserveColumnNames true/false. Default false.

This should be performed as an after-execution step, like the `RDFFormatter` step.  The reason for that is that when we introduce subqueries, dependent subquerie may want to access columns resulting from the child ones with the question mark.

patofp:
This PR add a flag to strip question marks on columns for Spark DataFrame output.

Closes #195

### DIFF
--- a/modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
+++ b/modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
@@ -8,6 +8,8 @@ import org.apache.spark.sql.Row
 
 import com.gsk.kg.sparqlparser.TestConfig
 
+import java.io.ByteArrayOutputStream
+
 import com.holdenkarau.spark.testing.DataFrameSuiteBase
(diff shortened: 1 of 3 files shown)


modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
```diff
@@ -78,6 +80,50 @@ class CompilerSpec
...
+          |+------+---------+
+          |
+          |""".stripMargin
+
+      result shouldBe a[Right[_, _]]
+
+      val outCapture = new ByteArrayOutputStream
+      Console.withOut(outCapture) {
+        result.right.get.show()
+      }
```
pepegar, line 122: good idea ;)

pepegar (approved):
LGTM!

PR #229, opened 2021-04-19
pepegar/issue185

### LINKED ISSUE #185: JoinBGPs optimization
Now that we're doing `GRAPH` pushdown optimizations, our queries are getting more and more with the shape `BGP <- JOIN -> BGP`.

These queries can be transformed into a single BGP, since there's an implicit join between triples of a BGP.  This way these queries can benefit from other optimizations such as BGP compaction.

pepegar:
This PR introduces the JoinBGPs optimization, which given a JOIN(BGP(t1), BGP(t2)), produces a BGP(t1 ++ t2).

This optimization is useful for aligning the graph for compaction.

close #185


patofp (approved):
LGTM!

PR #233, opened 2021-04-21
avoid RDF formatting by default

pepegar:


modules/engine/src/main/scala/com/gsk/kg/engine/Compiler.scala
```diff
@@ -62,23 +74,27 @@ object Compiler {
...
-  def staticAnalysis[T: Basis[DAG, *]]: Phase[T, T] =
+  private def staticAnalysis[T: Basis[DAG, *]]: Phase[T, T] =
     Analyzer.analyze
 
-  def rdfFormatter: Phase[DataFrame, DataFrame] = {
+  private def rdfFormatter: Phase[DataFrame, DataFrame] = {
     Kleisli[M, DataFrame, DataFrame] { inDf =>
       M.ask[Result, Config, Log, DataFrame].map { config =>
-        RdfFormatter.formatDataFrame(inDf, config)
+        if (config.formatRdfOutput) {
```
patofp, line 93: 👍

modules/engine/src/main/scala/com/gsk/kg/engine/Compiler.scala
```diff
@@ -25,13 +25,25 @@ object Compiler {
       .run(query)
       .runA(config, df)
 
+  // scalastyle:off
+  def explain(query: String)(implicit sc: SQLContext): Unit = {
```
patofp, line 29: Good idea!

patofp (approved):
LGTM!

PR #237, opened 2021-04-22
column based RdfFormatter

pepegar:
change RdfFormatter to a column based approach.

I've also added sbt-assembly to the build, but it's not actually being used, it's for ease of use when generating a fat jar to deploy


build.sbt
```diff
@@ -237,6 +237,12 @@ lazy val `bellman-spark-engine` = project
     }
     """
   )
+  .settings(
```
patofp, line 240: Yeah, this will be useful to test the Engine in standalone mode.

modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
```diff
@@ -1206,7 +1206,7 @@ class CompilerSpec
           result shouldBe a[Right[_, _]]
           result.right.get.collect.length shouldEqual 1
           result.right.get.collect.toSet shouldEqual Set(
-            Row("_:Henry", 21)
+            Row("_:Henry", "21")
```
patofp, line 1209: Could this quotation of non-strings cause unexpected output results on integration tests Bindu is doing?

patofp (approved):
LGTM! Just a comment to take a look at.

modules/engine/src/test/scala/com/gsk/kg/engine/CompilerSpec.scala
pepegar: It may, we need to test that.  but in any case this was against the expected schema we're using all the time of S, P and O having type String on result.

project/plugins.sbt
```diff
@@ -5,3 +5,4 @@ addSbtPlugin("com.geirsson"    % "sbt-ci-release"        % "1.5.7")
 addSbtPlugin("org.scalameta"   % "sbt-scalafmt"          % "2.4.2")
 addSbtPlugin("ch.epfl.scala"   % "sbt-scalafix"          % "0.9.27")
 addSbtPlugin("com.47deg"       % "sbt-microsites"        % "1.3.3")
+addSbtPlugin("com.eed3si9n"    % "sbt-assembly"          % "0.15.0")
```
patofp, line 8: Yeah, this will be useful to test the Engine in standalone mode.

patofp (approved):

PR #247, opened 2021-04-27
wrapped parser into engine Result monad

### LINKED ISSUE #231: Parser exceptions should be catched into a Result and mapped into a EngineError

patofp:
This PR wraps the engine method `QueryConstruct.parse` into a `Result` monad so it can be managed as the rest of the phases of the Engine. 

Now, Jena exceptions are captured into the left projection of `Result` so errors can be handled in an homogeneous way through all the engine.

Closes #231


geoffrey-pitman (approved):
Nice work!

PR #252, opened 2021-04-28
Support for regex function without flags

### LINKED ISSUE #128: Add support for REGEX without flags

geoffrey-pitman:
This PR adds the REGEX function without flags.  It also removes doc tables that are causing the CI job to fail.

Closes #128


patofp (commented):
GJ @geoffrey-pitman! Can you please add an issue to the backlog for **REGEX with flags**?

patofp (approved):

PR #254, opened 2021-04-28
added support for string on aggregates MIN and MAX

### LINKED ISSUE #232: Support strings in aggregate functions MAX, MIN
This query right now it is not supported due to MAX and  MIN functions only support `float` when they also support `strings` and others. Eg:

```
PREFIX peop: <http://people.example/>

SELECT ?y ?minName
WHERE {
  peop:alice peop:knows ?y .
  {
    SELECT ?y (MIN(?name) AS ?minName)
    WHERE {
      ?y peop:name ?name .
    } GROUP BY ?y
  }
}
```

patofp:
This PR adds support for strings on MIN and MAX aggregate functions. Out put has been contrasted with Apache Jena.

Closes #232


geoffrey-pitman (approved):
👍

PR #299, opened 2021-04-29
Support for regex with flags

### LINKED ISSUE #253: Add support for REGEX with flags
https://www.w3.org/TR/sparql11-query/#func-regex

https://www.w3.org/TR/xpath-functions/#flags

geoffrey-pitman:
This PR adds support for calling the SPARQL function REGEX with flags.

Closes #253


patofp (changes requested):
GJ @geoffrey-pitman ! I would add some test also to CompilerSpec that is acting as our integration tests.

patofp (approved):
Nice! LGTM!

PR #301, opened 2021-04-30
Support for order by

### LINKED ISSUE #89: Add support for ORDER BY

patofp:
This PR add support for queries with order by like this one:

```
PREFIX foaf:    <http://xmlns.com/foaf/0.1/>

SELECT ?name
WHERE { ?x foaf:name ?name ; foaf:age ?age }
ORDER BY DESC(?name) ?age DESC(?age) ASC(?name) DESC((isBlank(?x) || isBlank(?age)))
```

Where variables, functions and binary operations can be defined as conditions.

Closes #89

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
@@ -7,7 +7,7 @@ import cats.{Group => _, _}
 
 import higherkindness.droste.{Project => _, _}
 
-import com.gsk.kg.engine.DAG._
+import com.gsk.kg.engine.DAG.{Order => DAGOrder, _}
 import com.gsk.kg.engine.data.ChunkedList
 import com.gsk.kg.sparqlparser.StringVal.GRAPH_VARIABLE
 import com.gsk.kg.sparqlparser.StringVal.VARIABLE
@@ -85,6 +85,7 @@ object FindUnboundVariables {
       case Limit(limit, r)         => r.pure[ST]
       case Distinct(r)             => r.pure[ST]
       case Group(vars, func, r)    => r.pure[ST]
+      case DAGOrder(variable, r)   => r.pure[ST]
       case Noop(trace)             => Set.empty.pure[ST]
     }
 }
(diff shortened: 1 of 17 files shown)


geoffrey-pitman (approved):
👍

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
```diff
@@ -85,6 +85,7 @@ object FindUnboundVariables {
       case Limit(limit, r)         => r.pure[ST]
       case Distinct(r)             => r.pure[ST]
       case Group(vars, func, r)    => r.pure[ST]
+      case DAGOrder(variable, r)   => r.pure[ST]
```
pepegar, line 88: Shouldn't we be checking this `variable` to see if it's bound or not?

modules/parser/src/test/scala/com/gsk/kg/sparqlparser/OrderConditionParserSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+//(triple ?x <http://xmlns.com/foaf/0.1/name> ?name)
+//(triple ?x <http://xmlns.com/foaf/0.1/empId> ?emp)
+//)))
+//
+//(project (?name)
+//(order (?name ?emp (desc ?emp) (asc ?name) (desc (|| (isBlank ?x) (isBlank ?emp))))
+//(bgp
+//(triple ?x <http://xmlns.com/foaf/0.1/name> ?name)
+//(triple ?x <http://xmlns.com/foaf/0.1/empId> ?emp)
+//)))
```
pepegar, line 115:
```suggestion
```

pepegar (approved):
jUst added a couple of comments, but this looks great!

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
patofp: oh! that name is incorrect as that node is a list of conditions:

```
  @Lenses final case class Order[A](conds: NonEmptyList[ConditionOrder], r: A)
      extends DAG[A]
```

And ConditionOrder:

```
sealed trait ConditionOrder extends Expression
object ConditionOrder {
  final case class ASC(e: Expression)  extends ConditionOrder
  final case class DESC(e: Expression) extends ConditionOrder
}
```

So with that `e: Expression` I think it is already being checked whether are variables bounded. Is that correct? Although I'm going to check it.

modules/parser/src/test/scala/com/gsk/kg/sparqlparser/OrderConditionParserSpec.scala
patofp: upssss! 😅

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
patofp: Opened new issue to check variables on some nodes: #313

PR #316, opened 2021-05-04
Add expression variables to find unbounded

### LINKED ISSUE #313: Revisit FindUnboundedVariables for FILTER, GROUP and ORDER_BY nodes

patofp:
This PR adds static analysis support for unbounded variables on GROUP BY, ORDER BY and FILTER. 
When some variables on statement conditions are not defined in the WHERE or BGPs clauses, then it fails with an error.

Closes #313

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
@@ -81,11 +81,18 @@ object FindUnboundVariables {
           .flatMap(_ => Set.empty[VARIABLE].pure[ST])
       case LeftJoin(l, r, filters) => (l union r).pure[ST]
       case Union(l, r)             => (l union r).pure[ST]
-      case Filter(funcs, expr)     => expr.pure[ST]
-      case Join(l, r)              => r.pure[ST]
-      case Offset(offset, r)       => r.pure[ST]
-      case Limit(limit, r)         => r.pure[ST]
-      case Distinct(r)             => r.pure[ST]
+      case Filter(funcs, expr) =>
+        val funcsVars = funcs.toNes.toSortedSet
+          .foldLeft(Set.empty[VARIABLE]) { case (acc, func) =>
+            acc ++ FindVariablesOnExpression.apply[Expression](func)
+          }
+        for {
+          declared <- State.get
+        } yield (funcsVars diff declared) ++ expr
+      case Join(l, r)        => r.pure[ST]
+      case Offset(offset, r) => r.pure[ST]
+      case Limit(limit, r)   => r.pure[ST]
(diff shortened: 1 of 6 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
```diff
@@ -76,16 +78,79 @@ object FindUnboundVariables {
...
-      case Noop(trace)             => Set.empty.pure[ST]
+      case Filter(funcs, expr) =>
+        val funcsVars = funcs.toNes.toSortedSet
+          .foldLeft(Set.empty[VARIABLE]) { case (acc, func) =>
+            acc ++ FindVariablesOnExpression.apply[Expression](func)
+          }
+        for {
+          declared <- State.get
+        } yield (funcsVars diff declared) ++ expr
+      case Join(l, r)        => r.pure[ST]
```
pepegar, line 92: We should also fix this right?
```suggestion
      case Join(l, r)        => (l union r).pure[ST]
```

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
```diff
@@ -1,9 +1,17 @@
...
 
+import cats.kernel.Order
+
 /** @see Model after [[https://www.w3.org/TR/sparql11-query/#rExpression]]
   */
 sealed trait Expression
 
+object Expression {
+  implicit def order[T <: Expression]: Order[T] = new cats.Order[T] {
+    override def compare(x: T, y: T): Int = -1
```
pepegar, line 11: Where are we using this instance?  I think we should give a more useful one.  We could even use `Order.fromComparable` if needed

pepegar (approved):
Looks good to me, just added a couple of small comments

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
patofp: Yeah, didn't knew about that method, let me try it. Right now is being used in used in FindUnboundVarialbes:85, 101 in method `.toNes` that receives an implicit `Order[Expression]`.

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
patofp: Good catch!

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
patofp: Resolved, no need for that instance anymore.

geoffrey-pitman (approved):
Looks good!

PR #317, opened 2021-05-04
add logs to compiler

pepegar:
This PR adds logs to the different compiler phases.  It's not fully traced yet, but it adds a good start and infrastructure for easier logging.


patofp (approved):
Nice! LGTM

PR #326, opened 2021-05-05
accept angle brackets in input uris

### LINKED ISSUE #314: Accept angle brackets in input uris
We need to make our engine accept angle brackets in input dataframes so that no pre or post processing is needed.

pepegar:
closes #314


patofp (approved):
GJ!

PR #330, opened 2021-05-06
remove duplicates from Construct

### LINKED ISSUE #321: Remove duplicates from "Construct" query result
Since the result assumes to be a graph of triples. It makes sense to preform this step in the engine.

patofp:
This PR eliminates duplicate BGPs from CONSTRUCT queries.

This is congruent with Apache Jena that they also remove duplicates.

Closes #321


geoffrey-pitman (approved):
Looks good!

PR #331, opened 2021-05-06
STR on uris

### LINKED ISSUE #328: Calling STR() on an URI, should it drop angle brackets?
For example: 
```
PREFIX  dm:   <http://gsk-kg.rdip.gsk.com/dm/1.0/>
PREFIX  litn:  <http://lit-search-api/node/>
PREFIX  litp:  <http://lit-search-api/property/>

CONSTRUCT {
  ?Document a litn:Document .
  ?Document litp:docID ?docid .
}
WHERE{
  ?d a dm:Document .
  BIND(STRAFTER(str(?d), "#") as ?docid) .
  BIND(URI(CONCAT("http://lit-search-api/node/doc#", ?docid)) as ?Document) .
}
```
The result currently returns Document with double angle brackets ">>" at the end while docID has extra ending ">'".

pepegar:
What we're doing now is actually implementing STR as a function in the Func module, and stripping angle brackets off of uris

Closes #328


geoffrey-pitman (approved):
Nice! I like the trim function.  I didn't realize you could supply a value for it.  I'm going to use that for concat to strip away the quotes.

PR #334, opened 2021-05-06
don't orderBy by default

pepegar:
Avoid calling orderby on all constructs.  This will avoid performance bottlenecks, and the user can still orderby if they want by specifying it on the query

Closes to #333


geoffrey-pitman (approved):
Looks good!

PR #335, opened 2021-05-06
Support for strlen

### LINKED ISSUE #270: Add support for STRLEN string function
https://www.w3.org/TR/sparql11-query/#func-strings

geoffrey-pitman:
This PR adds support for the SparQL strlen function.

Closes #270


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/StrlenSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+    extends AnyWordSpec
+    with Matchers
+    with DataFrameSuiteBase
+    with TestConfig {
+
+  import sqlContext.implicits._
+
+  "perform STRLEN function correctly" when {
+
+    "used on string literals" in {
```
patofp, line 22: I think we should start to add tests on URIs and not only on string literals for string functions because of <>. What do you think @pepegar @geoffrey-pitman?

In this case image `?name` is a URI instead of a string literal. Should we count the angle brackets as a countable character for strlen? 
Are we counting double quotes as contable also?

I think we should contrast to Apache Jena.

patofp (changes requested):

geoffrey-pitman:
String type parsing needs more investigation before adding more string functions.

PR #339, opened 2021-05-07
fix uri function

### LINKED ISSUE #337: URI function not outputting URI with angle brackets
PREFIX dm: <http://gsk-kg.rdip.gsk.com/dm/1.0/>
SELECT ?o ?re
FROM <http://gsk-kg.rdip.gsk.com/umls/>
{
	?s dm:stringLiteral ?o
	BIND(URI("http://test.com") as ?re)
}

image

pepegar:
Since uris didn't need to contain angle brackets before, this function
was just a simple identity.  Now we're actually checking if the
received value is a angle bracket wrapped uri, or if it's a string
that needs to be wrapped.

Closes #337


patofp (approved):
LGTM

PR #348, opened 2021-05-11
Make aggregations work on typed literals

### LINKED ISSUE #341: make sure aggregations work on typed literals
Aggregations such as AVG, SUM, etc should work correctly not only on columns containing literal numeric values, but also on typed literals (`"1.0"^^xsd:float`).

the approach we'll follow is similar to Jena's.  We'll only accept rdf literal operandsss with types integer, float, decimal, and double.  We'll error otherwise.

As per return type, each function will decide their correct output type based on the nature of the function itself, and the operand types.

pepegar:
For numeric aggregations we're going to follow Jena's example and only allow RDF literals annotated with a numeric type from XSD.  Users will be able to use `strdt` if they want to convert _raw_ numeric values to RDF literal ones.

Closes #341


patofp (approved):
LGTM

PR #351, opened 2021-05-12
Rdf tests

### LINKED ISSUE #197: Look into implementing RDF-Tests
Could be useful for testing our Sparql coverage and missing features.

http://w3c.github.io/rdf-tests/

ideally we would publish it to end users so they know what features they can use

### LINKED ISSUE #332: Ensure errors are returned to the user
currently we're doing .right.get on the `df.sparql` function, so if we get a left containing some errors, they get ignored.

Let's think of a way of returning these to the user.

pepegar:
this PR introduces rdf-tests suite.  Currently I'm not hooking rdf-tests to CI, because our build wouldn't pass :)

However, we can use `sbt rdf-tests/test` to see what features we're supporting, what queries work and don't work, etc.  I've also tweaked `df.sparql` syntax so that it shows errors to the user.

# The approach

What I'm doing in the approach is separating between different possible scenarios. 
- If the test passes, good, ✔️ 
- if an `EngineError` is raised, the test is cancelled.  This makes it easier for spotting non-supported features.
- Otherwise, the test fails

Closes #197
Closes #332


modules/rdf-tests/src/test/scala/com/gsk/kg/RdfTests.scala
```diff
@@ -0,0 +1,178 @@
...
+import org.apache.spark.sql.Column
+import org.apache.jena.query.ResultSetFactory
+import org.apache.spark.sql.DataFrame
+import syntax._
+import scala.util.Try
+import scala.util.Failure
+import scala.util.Success
+import com.gsk.kg.sparqlparser.EngineError
+
+class RdfTests extends AnyWordSpec with Matchers with DataFrameSuiteBase {
```
pepegar, line 26: In case it makes it easier to find it, this is the file that implements the tests :)

patofp (approved):
Wow more than 1000 files. I've only take a look at build.sbt and syntax package. 😆 
LGTM!

patofp (approved):
LGTM! The git submodule makes sense. GJ!

PR #352, opened 2021-05-12
Add coverage strafter

### LINKED ISSUE #340: Add test coverage on STRAFTER
Add test coverage for all this cases: https://www.w3.org/TR/n-triples/#sec-literals and URIs also

patofp:
This PR adds test coverage to STRAFTER on edge cases, specially on typed and localized strings.

It also sets a way to add same tests to other string functions as STRBEFORE, STRSTARTS, STREND and CONTAINS.

Closes #340

### DIFF
--- a/modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
+++ b/modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
@@ -78,15 +78,21 @@ sealed trait StringVal extends StringLike {
   }
 }
 object StringVal {
-  final case class STRING(s: String)                   extends StringVal
-  final case class DT_STRING(s: String, tag: String)   extends StringVal
+  final case class STRING(s: String)                 extends StringVal
+  final case class DT_STRING(s: String, tag: String) extends StringVal
+  object DT_STRING {
+    def unApply(t: DT_STRING): String = s""""${t.s}"^^${t.tag}"""
+  }
   final case class LANG_STRING(s: String, tag: String) extends StringVal
-  final case class NUM(s: String)                      extends StringVal
-  final case class VARIABLE(s: String)                 extends StringVal
-  final case object GRAPH_VARIABLE                     extends StringVal { val s = "*g" }
-  final case class URIVAL(s: String)                   extends StringVal
-  final case class BLANK(s: String)                    extends StringVal
-  final case class BOOL(s: String)                     extends StringVal
+  object LANG_STRING {
+    def unApply(l: LANG_STRING): String = s""""${l.s}"${l.tag}"""
+  }
+  final case class NUM(s: String)      extends StringVal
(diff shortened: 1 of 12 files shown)


modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
```diff
@@ -78,15 +78,21 @@ sealed trait StringVal extends StringLike {
   }
 }
 object StringVal {
-  final case class STRING(s: String)                   extends StringVal
-  final case class DT_STRING(s: String, tag: String)   extends StringVal
+  final case class STRING(s: String)                 extends StringVal
+  final case class DT_STRING(s: String, tag: String) extends StringVal
+  object DT_STRING {
+    def unApply(t: DT_STRING): String = s""""${t.s}"^^${t.tag}"""
```
pepegar, line 84: I'd rename this to `toString` for example, unapply is the method used under the hood for pattern matching.

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
```diff
@@ -78,15 +78,21 @@ sealed trait StringVal extends StringLike {
...
+  }
   final case class LANG_STRING(s: String, tag: String) extends StringVal
-  final case class NUM(s: String)                      extends StringVal
-  final case class VARIABLE(s: String)                 extends StringVal
-  final case object GRAPH_VARIABLE                     extends StringVal { val s = "*g" }
-  final case class URIVAL(s: String)                   extends StringVal
-  final case class BLANK(s: String)                    extends StringVal
-  final case class BOOL(s: String)                     extends StringVal
+  object LANG_STRING {
+    def unApply(l: LANG_STRING): String = s""""${l.s}"${l.tag}"""
```
pepegar, line 88: Same here.  Also, shouldn't we use @ here?
```suggestion
    def unApply(l: LANG_STRING): String = s""""${l.s}"@${l.tag}"""
```

pepegar (approved):
Added a couple of comments.  Looks good otherwise!

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
pepegar: or maybe `toRdfLiteral`?

patofp: Yeah, that name seems more appropriate! 👍

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
patofp: Hmm, I'll take a look on this.

PR #355, opened 2021-05-13
Support for strbefore

### LINKED ISSUE #269: Add support for STRBEFORE string function
https://www.w3.org/TR/sparql11-query/#func-strings

geoffrey-pitman:
This PR fixes the STRBEFORE function, and now properly handles cases between simple strings, localized strings, and typed strings.  It also fixes a typo in the ToTree file, where REGEX was erroneously used in place of SUBSTR.  Additionally, the CONCAT function now uses trim instead of a regex replace.

Closes #269


patofp (approved):
LGTM!

PR #370, opened 2021-05-18
make FindUnboundVariables find unbound vars in nested queries

### LINKED ISSUE #234: Add static analysis on subqueries for inner variable bindings
There are some queries which may fail due to some variable bindings issues and can be detected previous to execution. Eg:

```
SELECT ?y ?name
WHERE {
  ?y foaf:knows ?x .
  {
    SELECT ?x
    WHERE {
      ?x foaf:name ?name .
    }
  }
}
```
 
In this query, in the inner subquery on the `SELECT` statement the variable `?name` is missing, but in the outer query then it binding, this produces an error because the output dataframe from the inner query lacks the column `?name` so it can't be binded afterwards when the outer query is executed. To do so the query must look like this:

```
SELECT ?y ?name
WHERE {
  ?y foaf:knows ?x .
  {
    SELECT ?x ?name // <--- Variable ?name have been added to SELECT statement
    WHERE {
      ?x foaf:name ?name .
    }
  }
}
```

This can be analyzed statically before execution so an error can be thrown to the user.

pepegar:
In this PR I needed to change the implementation of FindUnboundVariables.  the reason for that is that previously we were using a State monad for containing the declared vars, but that's not correct.  It's not correct because a piece of a query `a` may declare a variable and put it in the state, and a piece `b` may find it there, even if `b` shouldn't find that var in their lexical scope.

The approach we're following now is to have a simple cata that carries two values in a tuple, the declared variables, and the unbound variables.

closes #234

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
@@ -130,6 +130,8 @@ object FindVariablesOnExpression {
         case MAX(e)                          => e
         case AVG(e)                          => e
         case SAMPLE(e)                       => e
+        case LCASE(e)                        => e
+        case UCASE(e)                        => e
         case GROUP_CONCAT(e, separator)      => e
         case STRING(s)                       => Set.empty[VARIABLE]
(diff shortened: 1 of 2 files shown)


patofp (approved):
LGTM! GJ @pepegar

PR #375, opened 2021-05-18
Add support for language and typed strings

### LINKED ISSUE #372: Fix STRSTARTS string function
Add support for language tags, typed strings, and quoted strings.

geoffrey-pitman:
This PR adds support for language and type compatibility for the STRSTARTS function and fixes the issue of including quotes in the substring index.

Closes #372


modules/engine/src/main/scala/com/gsk/kg/engine/Func.scala
```diff
@@ -302,8 +302,23 @@ object Func {
...
+    when(
+      col.contains("\"@"),
+      trim(substring_index(col, "\"@", 1), "\"").startsWith(s)
+    )
+      .when(
+        col.contains("\"^^"),
+        trim(substring_index(col, "\"^^", 1), "\"").startsWith(s)
+      )
+      .otherwise(trim(col, "\"").startsWith(s))
+  }
```
pepegar, line 321: We're using the same code for extracting the underlying string in `strends` (see lines 285-293).  Could you abstract them to a single method?

pepegar (approved):
LGTM, just added a minor comment

modules/engine/src/main/scala/com/gsk/kg/engine/Func.scala
geoffrey-pitman: @pepegar sure thing!
https://github.com/gsk-aiops/bellman/blob/fix-strstarts/modules/engine/src/main/scala/com/gsk/kg/engine/Func.scala#L278-L309

PR #407, opened 2021-06-02
Minus

pepegar:


modules/parser/src/main/scala/com/gsk/kg/sparqlparser/ExprParser.scala
```diff
@@ -191,13 +198,15 @@ object ExprParser {
...
         | unionParen
+        | minusParen
         | extendParen
         | filterSingleParen
         | filterListParen
         | groupParen
         | orderParen
         | tableParen
         | existsParen
+        | minusParen
```
patofp, line 209: I think this one is duplicated

patofp (changes requested):

patofp (approved):
LGTM!

PR #408, opened 2021-06-02
Add support in/not in

### LINKED ISSUE #261: Add support for NOT IN function

patofp:
This PR adds support for IN/NOT IN function. E.g:

```

SELECT ?title
WHERE {
 ?film foaf:title ?title .
 FILTER (lang(?title) NOT IN ("en", "es"))
}
```

Closes #261

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncForms.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncForms.scala
@@ -1,8 +1,8 @@
 package com.gsk.kg.engine.functions
 
 import org.apache.spark.sql.Column
+import org.apache.spark.sql.functions.lit
 import org.apache.spark.sql.functions.not
-
 import com.gsk.kg.engine.functions.Literals.DateLiteral
 import com.gsk.kg.engine.functions.Literals.promoteNumericBoolean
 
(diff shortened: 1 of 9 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncForms.scala
```diff
@@ -96,4 +97,28 @@ object FuncForms {
...
+    * @return
+    */
+  def in(e: Column, xs: List[Column]): Column =
+    xs.foldLeft(lit(false)) { case (acc, x) =>
+      acc || {
+        val xType    = x.expr.dataType
+        val castExpr = e.cast(xType)
+        when(
+          castExpr.isNotNull,
+          castExpr.contains(x)
```
pepegar, line 121: can this `contains` return false positives, like in `castExpr = lit("hello")` and `x = lit("hell")`

patofp: 🤦 , good catch @pepegar. Also I updated the function to take into account when some expression has an error `lit(null)`. E.g: 
`2 IN (3, 1/0)	raises an error`

pepegar (approved):

PR #411, opened 2021-06-04
fixed quotation on plain literal already quotated

### LINKED ISSUE #393: extra double quotes in select query result
The object literal result has extra "" around it.

patofp:
This PR fixes quotation when plain literals are already quoted so extra double quotes won't be added.

Closes #393


pepegar (approved):
LGTM

PR #416, opened 2021-06-07
Query extractor

pepegar:
This PR introduces a new QueryExtractor object.

`QueryExtractor` has a `extractInfo` method that receives a query as a string and returns a tuple with a map of all query params related to their respective graph URIs, and a _clean_ query.  The _clean_ query has all the graph URIs stripped from query parameters.

Closes AIPL-3209

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
@@ -20,7 +20,10 @@ object QueryExtractor {
   final case class QueryParam(param: String, value: String)
 
   def extractInfo(q: String): (String, Map[String, List[QueryParam]]) = {
-    val (query, graphs) = QueryConstruct.parse(q, Config.default).right.get
+    val (query, graphs) = QueryConstruct.parse(q, Config.default) match {
+      case Left(a) => throw new Exception(a.toString)
+      case Right(b) => b
+    }
 
     (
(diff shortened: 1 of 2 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
```diff
@@ -0,0 +1,239 @@
...
+      case ProjectF(vars, r)     => r
+      case QuadF(s, p, o, g)     => s"QuadF(s, p, o, g)"
+      case DistinctF(r)          => s"$r DISTINCT"
+      case GroupF(vars, func, r) => s"$r GROUP BY ${printVars(vars)}"
+      case OrderF(conds, r) =>
+        s"$r ORDER BY ${conds.asInstanceOf[Seq[Expression]].map(printExpression).mkString(" ")}"
+      case OffsetLimitF(None, None, r)       => s"OffsetLimitF(None, None, r)"
+      case OffsetLimitF(None, Some(l), r)    => s"OffsetLimitF(None, Some(l)"
+      case OffsetLimitF(Some(o), None, r)    => s"OffsetLimitF(Some(o)"
+      case OffsetLimitF(Some(o), Some(l), r) => s"OffsetLimitF(Some(o)"
```
patofp, line 217: I think there is a missing closing `)` on this nodes.

patofp (approved):
LGTM!

PR #418, opened 2021-06-08
Add support if

### LINKED ISSUE #257: Add support for IF function

patofp:
This PR adds support for IF function. E.g:

```

SELECT ?name ?cat
WHERE {
 ?s foaf:name ?name .
 ?s foaf:age ?age .
 BIND(IF(?age < 18, "child", "adult") AS ?cat)
}
```

It also implements a function for [Effective Boolean Value](https://www.w3.org/TR/sparql11-query/#ebv).

Closes #257

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncForms.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncForms.scala
@@ -205,11 +205,16 @@ object FuncForms {
     * @param ifFalse
     * @return
     */
-  def `if`(cnd: Column, ifTrue: Column, ifFalse: Column): Column =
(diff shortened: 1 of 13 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncForms.scala
```diff
@@ -195,4 +195,104 @@ object FuncForms {
...
+      isNumericLiteral(col),
+      ifNumericLit
+    ).when(
+      typed.tag === lit(""),
+      ifPlainLit
+    ).when(
+      isStringLiteral(col),
+      ifStringLit
+    ).otherwise(nullLiteral)
+  }
```
pepegar, line 297: 😮 👍🏻

pepegar (approved):

PR #422, opened 2021-06-09
Fix literals return no results

patofp:
This PR fixes BGP matching on literals:

- String literals
- Boolean literals


pepegar (approved):
👏🏻 👏🏻 Thanks mate

PR #435, opened 2021-06-17
Add UUID function

JuanMGN:
Add uuid support function according specification, https://www.w3.org/TR/sparql11-query/#func-uuid

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
@@ -197,6 +197,7 @@ object QueryExtractor {
       case ExpressionF.BOOL(s)     => s
       case ASC(e)                  => s"ASC($e)"
       case DESC(e)                 => s"DESC($e)"
+      case UUID()                  => "UUID()"
     }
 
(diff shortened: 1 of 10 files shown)


pepegar:
Seems that the files are not properly formatted.   We use scalafmt, scalafix, and scalastyle in CI to test that the style is correct.

You can fix this by running `sbt lint` directly on the repository :)

modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
```diff
@@ -253,6 +253,7 @@ object QueryExtractor {
       case ExpressionF.BOOL(s)             => s
       case ASC(e)                          => s"(asc $e)"
       case DESC(e)                         => s"(desc $e)"
+      case UUID()                          => "UUID()"
```
JuanMGN, line 256: I'm not sure if this change is neccessary, may I have to write (UUID) ??

pepegar: Yep, it should be `(uuid)`

pepegar (approved):
Apart of the QueryExtractor comment, looks good to me!

PR #437, opened 2021-06-18
Add ceil function

JuanMGN:
Add ceil function according the specification: https://www.w3.org/TR/sparql11-query/#func-ceil

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncNumerics.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncNumerics.scala
@@ -1,6 +1,7 @@
 package com.gsk.kg.engine.functions
 
 import org.apache.spark.sql.Column
+import org.apache.spark.sql.functions.{ceil => sCeil}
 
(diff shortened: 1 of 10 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncNumerics.scala
```diff
@@ -23,7 +24,7 @@ object FuncNumerics {
     * @param col
     * @return
     */
-  def ceil(col: Column): Column = ???
+  def ceil(col: Column): Column = sCeil(col)
```
patofp, line 27: This can be a first implementation for plain numeric literals, but we should take into account typed numerical literals as: xsd:int, xsd:double, xsd:float, xsd:decimal.

I suggest opening a new ticket to support typed numerical literals and using functions from Literals.scala to implement it, also add test coverage on FuncNumericsSpec.scala.

patofp (changes requested):
I left two comments. For the rest LGTM!

patofp (approved):
LGTM!

PR #444, opened 2021-06-23
Add round function

JuanMGN:
Add round function according the specification: https://www.w3.org/TR/sparql11-query/#func-round


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncNumerics.scala
```diff
@@ -57,4 +49,25 @@ object FuncNumerics {
     * @return
     */
   def rand: Column = ???
+
+  private def apply(f: Column => Column, col: Column): Column =
```
patofp, line 53: 👍

patofp (approved):
LGTM!

pepegar (approved):

PR #445, opened 2021-06-24
fixed equals on plain string literals

patofp:
This PR fixes some issues when equals on plain strings that not match when extra double quoted.


pepegar (approved):
LGTM!

PR #446, opened 2021-06-24
Reorder BGPs optimization

pepegar:
This PR introduces a new optimization called reorder BGPs.  The ide
behind it is to avoid some cross joins that may occur when joining the
triples of a query.

Imagine the following three triples:

```sparql
?a :knows ?b
?z :knows ?y
?a :knows ?y
```

The first triple that bellman will execute in this case is `?a :knows
?b`, and then `?z :knows ?y`.  Since they don't share any variables,
spark will perform a cross join between them.  Then, the resulting
four column (`?a, ?b, ?z, ?y`) dataframe is joined with the two column
dataframe `?a, ?y`, performing an inner join on `?y`.

However the cross join could be avoided by reordering triples in this
way:

```sparql
?a :knows ?b
?a :knows ?y
?z :knows ?y
```

After this change, `?a` is shared between the first two triples, and
then `?y` between the second and the third, leaving us with two inner
joins.

## How this works

What we do in this optimization is to create a graph in which the
nodes are triples and edges exist when there are shared variables
between triples.

Then, we traverse the graph using BFS starting from the first triple,
and add the ones we visit to the list of triples inside the BGP,
leaving us a reordered BGP.


modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/ReorderBgps.scala
```diff
@@ -0,0 +1,108 @@
...
+          s"resulting query: ${result.toTree.drawTree}"
+        ),
+        ().pure[M]
+      ) *>
+      result.pure[M]
+  }
+
+  private def reorder(quads: ChunkedList[Expr.Quad]): ChunkedList[Expr.Quad] = {
+    val g = quads
+      .foldLeft(empty[Expr.Quad, Unit, Unit]) { (graph, quad) =>
```
patofp, line 55: I suppose here graph is being constructed.

patofp (approved):
LGTM! Nice job @pepegar

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/ReorderBgps.scala
pepegar: It is, yeah!  In quiver you use `Context` to add new nodes to the graph, and you pass to each context `inEdges`, `node`, `label`, `outEdges`.

patofp: Yeah I see, I'm taking a look at doc: https://verizon.github.io/quiver/

PR #447, opened 2021-06-25
Add rand function

JuanMGN:
Adding rand function according to specification
https://www.w3.org/TR/sparql11-query/#idp2130040


patofp (approved):
LGTM!

PR #449, opened 2021-06-25
Add abs function

JuanMGN:
Add abs function according specification:
https://www.w3.org/TR/sparql11-query/#func-abs


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncNumerics.scala
```diff
@@ -22,7 +23,7 @@ object FuncNumerics {
     * @param col
     * @return
     */
-  def abs(col: Column): Column = ???
+  def abs: Column => Column = col => apply(sAbs, col)
```
pepegar, line 26: Out of curiosity, why are you implementing this as a returned function instead of a method?

pepegar (approved):
LGTM

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncNumerics.scala
JuanMGN: Because It was more easy implementing the tests with these declaration. If you thing that is better to write how It was before, I can refactor the functions

pepegar: No need, I was just wondering about it. :)

PR #452, opened 2021-06-28
Add floor function

JuanMGN:
Add support to floor function according to specification:
https://www.w3.org/TR/sparql11-query/#func-floor


patofp (approved):
LGTM!

pepegar (approved):

PR #455, opened 2021-06-28
fix rdf-test to get more accuracy in reports

pepegar:
if using `cancel()` in rdf tests, we get too many false positives.  Let's change that with `fail()`


patofp (approved):
LGTM!

JuanMGN (approved):

PR #457, opened 2021-06-28
Add strUuid spark implementation

JuanMGN:
Add support to struuid function according the specification:
https://www.w3.org/TR/sparql11-query/#func-struuid


patofp (approved):
LGTM!

PR #461, opened 2021-06-29
Add now function

JuanMGN:
Add support to now function according the specification:
https://www.w3.org/TR/sparql11-query/#func-now


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/package.scala
```diff
@@ -30,4 +39,38 @@ package object compiler {
...
+    def eval(
+        df: DataFrame,
+        projection: Option[Column],
+        query: String,
+        expected: List[Row]
+    ): Assertion = {
+      val result = Compiler.compile(df, query, config)
+
+      val dfR: DataFrame = result match {
+        case Left(e)  => throw new Exception(e.toString)
```
patofp, line 59: Maybe instead of throwing a exception we can fail with something like:
`fail("I've got a bad feeling about this.")` or something similar: https://www.scalatest.org/user_guide/using_assertions

patofp (commented):
With the introduction of Evaluation we should think about adapting the rest of the tests to that same formula, if I'm not mistaken it would apply at least to function suites where a projection is normally used. This way test are as homogeneous as they can be across all the project. What are your thoughts?

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/package.scala
```diff
@@ -30,4 +39,38 @@ package object compiler {
...
 
+  object Evaluation
+      extends AnyWordSpec
+      with Matchers
+      with SparkSpec
+      with TestConfig {
+
+    val renamedColumn = "c1"
+
+    def eval(
```
pepegar, line 50: Can you add docs here explaining how should it be used?

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/DateParser.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import com.gsk.kg.sparqlparser.Date._
+
+import fastparse.MultiLineWhitespace._
+import fastparse._
+
+/** Functions on Functions on Dates and Times:
+  * https://www.w3.org/TR/sparql11-query/#func-date-time
+  */
+object DateParser {
```
pepegar, line 11: What do you think about renaming this to `DateTimeFuncsParser` or somethign like thatt? now it seems like it parses dates

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
```diff
@@ -152,3 +152,8 @@ object ConditionOrder {
   final case class ASC(e: Expression)  extends ConditionOrder
   final case class DESC(e: Expression) extends ConditionOrder
 }
+
+sealed trait Date extends Expression
+object Date {
```
pepegar, line 157: Same here, `DateTimeFunc` maybe?

pepegar (changes requested):

patofp (approved):
LGTM!

JuanMGN:
As Patricio says, I think adapting all the tests to this same formula is a good idea, if possible. Maybe, when Pepe finishes the ticket to improve the test time or during it.

pepegar (approved):

PR #463, opened 2021-06-30
Path properties algebra

patofp:
WIP: This PR adds support for path properties on parser side.

### DIFF
--- a/modules/parser/src/main/scala/com/gsk/kg/sparqlparser/PropertyExpression.scala
+++ b/modules/parser/src/main/scala/com/gsk/kg/sparqlparser/PropertyExpression.scala
@@ -5,8 +5,10 @@ sealed trait PropertyExpression
 final case class Alternative(pel: PropertyExpression, per: PropertyExpression)
     extends PropertyExpression
 final case class Reverse(e: PropertyExpression) extends PropertyExpression
-final case class SeqExpression(pel: PropertyExpression, per: PropertyExpression)
-    extends PropertyExpression
+final case class SeqExpression(
+    pel: PropertyExpression,
+    per: PropertyExpression
+)                                                  extends PropertyExpression
 final case class OneOrMore(e: PropertyExpression)  extends PropertyExpression
 final case class ZeroOrMore(e: PropertyExpression) extends PropertyExpression
 final case class ZeroOrOne(e: PropertyExpression)  extends PropertyExpression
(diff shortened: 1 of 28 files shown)


modules/parser/src/main/scala/com/gsk/kg/sparqlparser/PropertyExpression.scala
```diff
@@ -0,0 +1,23 @@
...
+    extends PropertyExpression
+final case class BetweenNAndM(n: Int, m: Int, e: PropertyExpression)
+    extends PropertyExpression
+final case class ExactlyN(n: Int, e: PropertyExpression)
+    extends PropertyExpression
+final case class NOrMore(n: Int, e: PropertyExpression)
+    extends PropertyExpression
+final case class BetweenZeroAndN(n: Int, e: PropertyExpression)
+    extends PropertyExpression
+final case class Uri(s: String) extends PropertyExpression
```
pepegar, line 23: Can you `@deriveFixedPoint` this? It will help us traversing the property path trees
```suggestion
import higherkindness.droste.macros.deriveFixedPoint

@deriveFixedPoint sealed trait PropertyExpression

final case class Alternative(pel: PropertyExpression, per: PropertyExpression)
    extends PropertyExpression
final case class Reverse(e: PropertyExpression) extends PropertyExpression
final case class SeqExpression(pel: PropertyExpression, per: PropertyExpression)
    extends PropertyExpression
final case class OneOrMore(e: PropertyExpression)  extends PropertyExpression
final case class ZeroOrMore(e: PropertyExpression) extends PropertyExpression
final case class ZeroOrOne(e: PropertyExpression)  extends PropertyExpression
final case class NotOneOf(es: List[PropertyExpression])
    extends PropertyExpression
final case class BetweenNAndM(n: Int, m: Int, e: PropertyExpression)
    extends PropertyExpression
final case class ExactlyN(n: Int, e: PropertyExpression)
    extends PropertyExpression
final case class NOrMore(n: Int, e: PropertyExpression)
    extends PropertyExpression
final case class BetweenZeroAndN(n: Int, e: PropertyExpression)
    extends PropertyExpression
final case class Uri(s: String) extends PropertyExpression
```

pepegar (approved):
LGTM!  Just added a minor comment for tthe future ;)

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/PropertyExpression.scala
patofp: Indeed, it is going to be needed for the next phase of generating a new algebra for property paths as we talked. 😉

PR #475, opened 2021-07-07
Add strlang function

### LINKED ISSUE #265: Add support for STRLANG function
https://www.w3.org/TR/sparql11-query/#func-strlang

puremachinery:
Adding support for the STRLANG function. https://www.w3.org/TR/sparql11-query/#func-strlang

Closes #265


puremachinery:
It looks like the check is failing on a linter warning for the BuiltInFuncParser now having over 70 methods? Shall I refactor BuiltInFuncParser, e.g. maybe pull abs/ceil/round/etc into a MathFuncParser?

patofp:
> It looks like the check is failing on a linter warning for the BuiltInFuncParser now having over 70 methods? Shall I refactor BuiltInFuncParser, e.g. maybe pull abs/ceil/round/etc into a MathFuncParser?

This is because scalastyle configuration, this is set here and can be modified:
https://github.com/gsk-aiops/bellman/blob/ea09e8a28e25ee530dacb306338984b0eb6a45b1/scalastyle-config.xml#L104

Said that, I think it is a good practice to keep things small, clear and modularized, so I think it is a good idea to split parsers into small ones, it makes them more human readable. 😆 

By the way, there is also a way turn off this style checks on particular code:

```
// scalastyle:off
...
// naughty stuff
...
// scalastyle:on
```

or:

`naughty()  // scalastyle:ignore`

[Scalastyle doc](http://www.scalastyle.org/configuration.html)

> I'll try that for now. (I also see the other warning about the Algebra, adding fix to QueryExtractor.)

Cool, whenever it is possible we try to be exhaustive in our pattern matchings. 😉

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/StrlangSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+    val query =
+      s"""
+         |PREFIX foaf: <http://xmlns.com/foaf/0.1/>
+         |SELECT ?strlang
+         |WHERE {
+         |?x foaf:name ?name .
+         |BIND(strlang(?name, $tag) as ?strlang)
+         |}""".stripMargin
+
+    print(query)
```
patofp, line 59: I think it is better not necessary to print things on test.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/StrlangSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+    (Compiler
+      .compile(df, query, config) match {
```
patofp, line 62: Here on tests we usually call the unsafe method `.right.get` on this Either, it will throw that exception. Although the correct way would be to instead of throwing that exception we would fail the test.

```
...
case Left(l) => fail(l.toString)
...
```

patofp (changes requested):
Just a couple of things to change on the tests, for the rest LGTM!

modules/engine/src/main/scala/com/gsk/kg/engine/ExpressionF.scala
```diff
@@ -205,11 +207,11 @@ object ExpressionF {
...
-      case BuiltInFunc.CEIL(s)                         => CEIL(s)
-      case BuiltInFunc.ROUND(s)                        => ROUND(s)
-      case BuiltInFunc.RAND()                          => RAND()
-      case BuiltInFunc.ABS(s)                          => ABS(s)
-      case BuiltInFunc.FLOOR(s)                        => FLOOR(s)
+      case MathFunc.CEIL(s)                            => CEIL(s)
+      case MathFunc.ROUND(s)                           => ROUND(s)
+      case MathFunc.RAND()                             => RAND()
+      case MathFunc.ABS(s)                             => ABS(s)
+      case MathFunc.FLOOR(s)                           => FLOOR(s)
```
pepegar, line 214: 👍🏻

pepegar (approved):
Looks good to me!

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/StrlangSpec.scala
puremachinery: Ah yes, did this while trying to verify something, but didn't mean to check it in.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/StrlangSpec.scala
puremachinery: Same as other comment, this was just for debugging, not meant to check in. Thanks.

puremachinery: Removed.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/StrlangSpec.scala
puremachinery: Removed.

patofp (approved):
LGTM!

PR #476, opened 2021-07-09
Add minutes function

JuanMGN:
# Description

Add minutes functions according specification:
https://www.w3.org/TR/sparql11-query/#func-minutes

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
@@ -98,6 +98,12 @@ object FuncDates {
       f(NumericLiteral(col).value)
     ).otherwise(nullLiteral)
 
+  /** Get hours, minutes of dateTime column
+    * @param col
+    * @param pos
+    * @return Column with
+    *         Integer if hours or minutes
+    */
   private def getTimeFromDateTimeCol(col: Column, pos: Int): Column = {
     val dateTimeRegex: String =
       "[0-9]{1,4}-[0-9]{1,2}-[0-9]{1,2}T[0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2}"
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
@@ -5,10 +5,14 @@ import org.apache.spark.sql.functions.current_timestamp
 import org.apache.spark.sql.functions.date_format
 import org.apache.spark.sql.functions.dayofmonth
 import org.apache.spark.sql.functions.format_string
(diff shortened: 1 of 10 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
```diff
@@ -17,8 +18,8 @@ import com.gsk.kg.engine.functions.Literals.nullLiteral
 
 object FuncDates {
 
-  val dateTimeRegex: String =
-    "[0-9]{1,4}-[0-9]{1,2}-[0-9]{1,2}T[0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2}"
+  private val Hours   = 3
```
patofp, line 21: Why are these defined globally? I can only see them used on their corresponding function scope.

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
```diff
@@ -101,4 +97,28 @@ object FuncDates {
...
+  /** Get hours, minutes of dateTime column
+    * @param col
+    * @param pos
+    * @return Column with
+    *         Integer if hours or minutes
+    */
+  private def getTimeFromDateTimeCol(col: Column, pos: Int): Column = {
+    val dateTimeRegex: String =
+      "[0-9]{1,4}-[0-9]{1,2}-[0-9]{1,2}T[0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2}"
+    val len = 2
```
patofp, line 110: Where is this variable being used?

patofp (changes requested):
Just a couple of things, for the rest LGTM!

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
JuanMGN: len is not used, I forgot delete it

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
JuanMGN: These variables( Minutes and Hours) are used when we call to function getTimeFromDateTimeCol. I think the code is more readable if the numbers are declared as a variable

patofp: Yeah, it is the scope of the variable what it should be changed. 😉

patofp (approved):
LGTM!

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
JuanMGN: done :)

PR #478, opened 2021-07-09
added implementation for seq property path

patofp:
# Description

This PR adds support for `seq` property path. E.g:

```

SELECT ?o
WHERE {
 ?s foaf:knows/foaf:knows/foaf:name ?o .
}
```

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
@@ -257,8 +257,7 @@ object Engine {
       Foldable[ChunkedList].fold(
         quads.mapChunks { chunk =>
           val condition = composedConditionFromChunk(df, chunk)
-          val a         = applyChunkToDf(chunk, condition, df)
-          a
+          applyChunkToDf(chunk, condition, df)
         }
       )
     }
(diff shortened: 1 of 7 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -237,7 +257,8 @@ object Engine {
       Foldable[ChunkedList].fold(
         quads.mapChunks { chunk =>
           val condition = composedConditionFromChunk(df, chunk)
-          applyChunkToDf(chunk, condition, df)
+          val a         = applyChunkToDf(chunk, condition, df)
+          a
```
pepegar, line 261: Nitpick: can we leave as it was? hehehe

modules/engine/src/main/scala/com/gsk/kg/engine/PropertyExpressionF.scala
```diff
@@ -15,35 +16,41 @@ import com.gsk.kg.sparqlparser.Result
 
 object PropertyExpressionF {
 
+  type ColOrDf = Either[Column, DataFrame]
```
pepegar, line 19: 👍🏻

pepegar (approved):
great job mate!

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
patofp: uuups! it was just for debug purpose 😆

PR #481, opened 2021-07-13
Add tz function

JuanMGN:
# Description

Add tz function according specification
https://www.w3.org/TR/sparql11-query/#func-tz

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
@@ -1,7 +1,6 @@
 package com.gsk.kg.engine.functions
 
 import org.apache.spark.sql.Column
-import org.apache.spark.sql.functions.concat
 import org.apache.spark.sql.functions.current_timestamp
 import org.apache.spark.sql.functions.date_format
 import org.apache.spark.sql.functions.dayofmonth
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
@@ -1,12 +1,15 @@
 package com.gsk.kg.engine.functions
 
 import org.apache.spark.sql.Column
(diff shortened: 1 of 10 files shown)


puremachinery:
LGTM, though I won't try to Approve since I'm still new to this codebase and prefer to have Pepe or Patricio also take a look before it gets auto-merged. :)

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
```diff
@@ -170,4 +148,97 @@ object FuncDates {
...
+
+  /** Get timezone of a datetime input in multiples formats
+    * @param col
+    * @return timezone formatted
+    *         usage
+    *         getTimeZoneComponents(2020-12-09T01:50:24.888Z) = Z
+    *         getTimeZoneComponents(2011-01-10T14:45:13.815-05:29) = -05:29
+    *         getTimeZoneComponents(2011-01-10T14:45:13.815+05:09) = 05:29
+    */
+  private def getTimeZoneComponents(col: Column): Column = {
```
pepegar, line 160: 👍🏻

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncDates.scala
```diff
@@ -170,4 +148,97 @@ object FuncDates {
...
+    * @return timezone formatted
+    *         usage
+    *         buildTimeZone(-05:29) = "\"-PT5H29M\"^^xsd:dateTime"
+    *         buildTimeZone(05:00)  = "\"PT5H\"^^xsd:dateTime"
+    */
+  private def buildTimeZone(
+      timeZone: Column,
+      PosSignOpt: Option[Int],
+      PosHours: Int,
+      PosMinutes: Int
```
pepegar, line 220: I'd convert these to lowercase.  They'd only be constants if they were plain variables, not arguments.

pepegar (approved):
LGTM

PR #482, opened 2021-07-15
testing out typed dataframes

pepegar:
# Description

Here's my attempt at a more typed core for Bellman.  I'm creating this PR as more documentation and conversation space than something I'd like to merge to `master`.

**TLDR;**: Our hypothesis in which, by parsing strings to actual values would improve time, didn't work.

I think using using a more typed approach (either with Frameless, or plain Datasets) makes sense for ETLs, if someone is moving data from point to point and performing transformations inside.  On the other hand if what you (we in this case) is to create a general purpose query engine, and the shape of data can change on every expression that the user adds to the query, we need to use the most genera API possible, `DataFrames` in our case.  Otherwise, the performance penalty described in **`Deserialization`** is just too high.

In the file `RdfValuesSpec` you can see the two simple tests I wanted to perform.  On one hand I'm just calling a function implemented for untyped columns, and then the sample functino implemented for `RdfValues`. 

The first one is **noticeably** faster, and has far less heap & metaspace fingerprint.

# Conclusion

I think we shouldn't go for this approach for now.  It won't solve the issue as we expected.

Some takeaways of this spike:

## 👍🏻 Types

our functions could be typed when using this approach. See `def strafter` in the diff.  It looks much better than working with Spark column ops, and is less error prone.

## 👎🏻 Deserialization

If we want to use the typed APIs of either `frameless.TypedDataset`, or `spark.sql.DataSet`, we need our functions to operate on values on the heap.  In order to do that, these operations **must** run outside of catalyst, and Catalyst needs to convert from Tungsten representation to JVM and back again.

I expected typed functions to do the opposite, be serialized to catalyst operations instead.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ lazy val Versions = Map(
   "scalatestplus"        -> "3.2.3.0",
   "sansa"                -> "0.7.1",
   "monocle"              -> "1.5.0",
+  "frameless"            -> "0.8.0",
   "discipline"           -> "1.1.2",
   "discipline-scalatest" -> "2.0.1",
   "reftree"              -> "1.4.0",
@@ -130,6 +131,7 @@ lazy val `bellman-spark-engine` = project
   .settings(
     libraryDependencies ++= Seq(
       "io.verizon.quiver"          %% "core"         % Versions("quiver"),
+      "org.typelevel"              %% "frameless-dataset" % Versions("frameless"),
       "org.apache.spark"           %% "spark-sql"    % Versions("spark") % Provided,
       "com.github.julien-truffaut" %% "monocle-core" % Versions("monocle"),
       "com.github.julien-truffaut" %% "monocle-macro" % Versions("monocle"),
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/rdf/RdfValues.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/rdf/RdfValues.scala
@@ -1,0 +1,47 @@
+package com.gsk.kg.engine.rdf
+
+import com.gsk.kg.engine.rdf.RdfValue._
+
+import frameless.Injection
+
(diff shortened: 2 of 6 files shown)


JNKHunter:
This is all part of fail fast! Good analysis and learnings.

patofp:
> # Description
> Here's my attempt at a more typed core for Bellman. I'm creating this PR as more documentation and conversation space than something I'd like to merge to `master`.
> 
> **TLDR;**: Our hypothesis in which, by parsing strings to actual values would improve time, didn't work.
> 
> I think using using a more typed approach (either with Frameless, or plain Datasets) makes sense for ETLs, if someone is moving data from point to point and performing transformations inside. On the other hand if what you (we in this case) is to create a general purpose query engine, and the shape of data can change on every expression that the user adds to the query, we need to use the most genera API possible, `DataFrames` in our case. Otherwise, the performance penalty described in **`Deserialization`** is just too high.
> 
> In the file `RdfValuesSpec` you can see the two simple tests I wanted to perform. On one hand I'm just calling a function implemented for untyped columns, and then the sample functino implemented for `RdfValues`.
> 
> The first one is **noticeably** faster, and has far less heap & metaspace fingerprint.
> 
> # Conclusion
> I think we shouldn't go for this approach for now. It won't solve the issue as we expected.
> 
> Some takeaways of this spike:
> 
> ## 👍🏻 Types
> our functions could be typed when using this approach. See `def strafter` in the diff. It looks much better than working with Spark column ops, and is less error prone.
> 
> ## 👎🏻 Deserialization
> If we want to use the typed APIs of either `frameless.TypedDataset`, or `spark.sql.DataSet`, we need our functions to operate on values on the heap. In order to do that, these operations **must** run outside of catalyst, and Catalyst needs to convert from Tungsten representation to JVM and back again.
> 
> I expected typed functions to do the opposite, be serialized to catalyst operations instead.

PR #484, opened 2021-07-19
Refactor sameTerms

JuanMGN:
apply some changes to sameTerms methods to fix the implementation

# Description

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/Literals.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/Literals.scala
@@ -1,5 +1,6 @@
 package com.gsk.kg.engine.functions
 
+import com.gsk.kg.engine.RdfFormatter
 import org.apache.spark.sql.Column
 import org.apache.spark.sql.functions.{concat => cc, _}
 import org.apache.spark.sql.types.StringType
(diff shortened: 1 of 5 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/functions/Literals.scala
```diff
@@ -2,12 +2,18 @@ package com.gsk.kg.engine.functions
...
+import org.apache.spark.sql.types.BooleanType
 import org.apache.spark.sql.types.StringType
 
+import com.gsk.kg.engine.RdfFormatter
+
 object Literals {
 
   // scalastyle:off
-  val nullLiteral = lit(null)
+  val nullLiteral: Column    = lit(null)
```
patofp, line 13: as this is also a constant I suggest refactoring to uppercase and maintain homogeneity.

```suggestion
  val NullLiteral: Column    = lit(null) // scalastyle:ignore
```

Also I think there is no need disable scalastyle on every constant as it only complains about `null`.

patofp (changes requested):

patofp (changes requested):
Just a little naming convention change requested. For the rest LGTM!

pepegar (approved):

patofp:
After analyzing the code it seems that the Typed Dataframe refactor fixed this problem, so closing this PR.

PR #485, opened 2021-07-19
add support one or more property path (+)

patofp:
# Description

This PR adds support for one or more length property path. E.g:

```

SELECT ?s ?o
WHERE {
 ?s foaf:knows+ ?o .
}
```

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
@@ -4,13 +4,8 @@ import cats.implicits.catsSyntaxEitherId
 import cats.syntax.either._
 
 import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.Encoder
 import org.apache.spark.sql.Row
-import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.functions._
-import org.apache.spark.sql.types.StringType
-import org.apache.spark.sql.types.StructField
-import org.apache.spark.sql.types.StructType
 
 import com.gsk.kg.engine.PropertyExpressionF.ColOrDf
 import com.gsk.kg.engine.functions.FuncForms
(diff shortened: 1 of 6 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
```diff
@@ -91,6 +91,157 @@ object FuncProperty {
...
+
+    implicit val spoEncoder: Encoder[Row] = RowEncoder(
+      StructType(
+        List(
+          StructField("s", StringType),
+          StructField("p", StringType),
+          StructField("o", StringType)
+        )
+      )
+    )
```
pepegar, line 226: Don't we have this very same encoder somewhere else?  I think we should extract it to a common place

modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
```diff
@@ -91,6 +91,157 @@ object FuncProperty {
...
+    * more triples to connect.
+    * @param baseDf Base dataframe with all the triples unconnected
+    * @param accDf The dataframe which it will accumulate every depth level of paths
+    * @param i Iteration index
+    * @param sIdx Subject index (for column renaming)
+    * @param pIdx Predicate index (for column renaming)
+    * @param oIdx Object index (for column renaming
+    * @return Dataframe with all the triples connected creating paths
+    */
+  private def traverse(
```
pepegar, line 138: What do you think about changing the name to something more specific?

pepegar (approved):
Added a  couple of cosmetic comments, but this looks amazing!

modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
patofp: Yeah, good idea!

PR #487, opened 2021-07-20
add support zero or more property path (*)

patofp:
# Description

This PR adds support for zero or more property path. E.g:

```

SELECT ?s ?o
WHERE {
 ?s foaf:knows* ?o .
}
```

## Type of change

- `bug` if it fixes a bug

Merge after #485

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
@@ -3,12 +3,14 @@ package com.gsk.kg.engine.properties
 import cats.implicits.catsSyntaxEitherId
 import cats.syntax.either._
 
+import org.apache.spark.sql.Column
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.functions._
 
 import com.gsk.kg.engine.PropertyExpressionF.ColOrDf
 import com.gsk.kg.engine.functions.FuncForms
+import com.gsk.kg.engine.functions.Literals.nullLiteral
 import com.gsk.kg.sparqlparser.EngineError
 import com.gsk.kg.sparqlparser.Result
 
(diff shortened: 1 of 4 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
```diff
@@ -181,6 +206,21 @@ object FuncProperty {
     }
   }
 
+  /** It returns a dataframe with all the vertex that points to itselft (0 length path), the predicate column will
```
pepegar, line 209:
```suggestion
  /** It returns a dataframe with all the vertices that point to itself (0 length path), the predicate column will
```

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/PropertyPathsSpec.scala
```diff
@@ -214,7 +217,15 @@ class PropertyPathsSpec
 
         val result = Compiler.compile(df, query, config)
 
-        result.right.get.collect.toSet shouldEqual Set()
+        result.right.get.collect.toSet shouldEqual Set(
+          Row("\"Charles\"", "\"Charles\""),
```
pepegar, line 221: Why do we get the this particular result?

modules/engine/src/test/scala/com/gsk/kg/engine/properties/FuncPropertySpec.scala
```diff
@@ -190,6 +190,74 @@ class FuncPropertySpec
...
+          Row(
+            "<http://example.org/Charles>",
+            null,
+            "<http://example.org/Charles>"
+          ),
+          Row(
+            "\"Charles\"",
+            null,
+            "\"Charles\""
+          ),
```
pepegar, line 242: Same here... Why does this row appear in the result set?  `"Charles"` only appears as `o` in the `<Charles> -> name -> "Charles"` triple.  doesn't have anything to do with `knows`

pepegar (changes requested):
Maybe I misunderstood the property path, but there are some results I don't really understand.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/PropertyPathsSpec.scala
patofp: Yeah, that's a good question. What is a path of length 0 in a graph?

First let's see what is a path: "A path is a trail in which all vertices (and therefore also all edges) are distinct."

If the trail only has one edge we say it has length 1, if it has two edges then is length 2 and therefore.

So what is the length 0 path? It would be the path from a vertex to itself and because they are the same vertex they can't be of length 1.

Every vertex in the graph has one and only one path 0 length that is the one to itself, in our case we consider everything in columns **s** and **o** as vertices and **p** as the edges. That solution is the representation of the 0 path length for the vertex named "Charles".

I've also contrasted with Jena and it also returns that solution.

pepegar: Talked via slack, resolved!

pepegar (approved):

modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala
patofp: Done!

PR #488, opened 2021-07-20
Add datatype function

### LINKED ISSUE #263: Add support for DATATYPE function

puremachinery:
Implementation of the datatype function.

```
iri  DATATYPE (literal literal)
Returns the datatype IRI of a literal.

If the literal is a typed literal, return the datatype IRI.
If the literal is a simple literal, return xsd:string
If the literal is literal with a language tag, return rdf:langString
```
More at https://www.w3.org/TR/sparql11-query/#func-datatype

Closes #263

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
@@ -108,7 +108,7 @@ object FuncTerms {
         when( // if the literal is a simple literal, return xsd:string
           isLiteral(col),
           "xsd:string"
-        ).otherwise( // not a literal. not defined by w3 spec. return lit("") or fail/exception/other?
+        ).otherwise( // input not a literal; not defined by w3 spec
           lit("")
         )
       )
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
@@ -3,9 +3,10 @@ package com.gsk.kg.engine.functions
 import org.apache.spark.sql.Column
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions.{concat => cc, _}
+
 import com.gsk.kg.engine.functions.Literals.TypedLiteral
-import com.gsk.kg.engine.functions.Literals.nullLiteral
 import com.gsk.kg.engine.functions.Literals.TypedLiteral.isTypedLiteral
+import com.gsk.kg.engine.functions.Literals.nullLiteral
 
 object FuncTerms {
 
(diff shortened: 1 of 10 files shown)


puremachinery:
Especially interested in eyes on the datatype function itself & whether the return given an invalid (non-literal) input is handled the way we want, since I'm unclear on overall project policy regarding invalid inputs & returning something empty vs treating it more explicitly as an error or exception.

https://github.com/gsk-aiops/bellman/pull/488/files#diff-a958d3b9fbb6126ebc06429e354a157e689a0a348ecf2247026f9d8bb021bcc4R99-R115

patofp:
> Especially interested in eyes on the datatype function itself & whether the return given an invalid (non-literal) input is handled the way we want, since I'm unclear on overall project policy regarding invalid inputs & returning something empty vs treating it more explicitly as an error or exception. @pepegar @patofp
> 
> https://github.com/gsk-aiops/bellman/pull/488/files#diff-a958d3b9fbb6126ebc06429e354a157e689a0a348ecf2247026f9d8bb021bcc4R99-R115

That's a good point, so right now when in the Sparql Spec says that in a particular case it should return an error what we are doing is returning `lit(null)`, and this is arguable if it is the best approach but I think we should stick on this for now to maintain homogeneity throughout all functions.

The thing is that you can't or at least I don't know how to force Spark driver to generate an exception and I'm not sure if that would be the best solution. 

Another option will be to create a special string that represents an error or if we had Dataset instead of Dataframe we would have a case object for example, but these also will propagate to all functions where we should do some error handling. And here we don't have the help of the type system, as we do have in Scala to force de developer to handle that cases, but instead it is part of the business logic but it is easy to forget to handle and it is error prone. Also this could lead into RDF not compilant if not well handled

The lit("") is another option but there could be some case where the user creates a query where he/she would like to match on empty string literals so it could lead into wrong number of solutions.

You can use this one: https://github.com/gsk-aiops/bellman/blob/f6b857562f3f896b2ff4ebbc0c16ecceb60504cd/modules/engine/src/main/scala/com/gsk/kg/engine/functions/Literals.scala#L10 so linter won't complain.

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
```diff
@@ -89,6 +90,30 @@ object FuncTerms {
...
+    ).otherwise(
+      when( // if the literal has a language tag, return rdf:langString
+        !lang(col).equalTo(lit("")),
+        "rdf:langString"
+      ).otherwise(
+        when( // if the literal is a simple literal, return xsd:string
+          isLiteral(col),
+          "xsd:string"
+        ).otherwise( // input not a literal; not defined by w3 spec
+          lit("")
```
patofp, line 112: as discussed here: https://github.com/gsk-aiops/bellman/pull/488#issuecomment-883533500 this is the case where we should use `nullLiteral` instead.

patofp (changes requested):
Just the change requested for the error case, for the rest LGTM!

pepegar (approved):

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
puremachinery: Why return null rather than the usual Scala Option/Some/None pattern?

puremachinery: Updated.

patofp (approved):
GJ!

PR #489, opened 2021-07-20
Add group concat function

JuanMGN:
# Description

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncAgg.scala
```diff
@@ -106,7 +106,6 @@ object FuncAgg {
     )
   }
 
-  // TODO: Implement group-concat https://github.com/gsk-aiops/bellman/issues/202
   def groupConcat(col: Column, separator: String): Column =
-    ???
+    array_join(collect_list(col), separator)
```
pepegar, line 110: Lovely @JuanMGN !  I thought we would need to use an UDF at this point!

pepegar (approved):
GJ!

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/AggregateParser.scala
```diff
@@ -27,14 +27,13 @@ object AggregateParser {
...
+          _ != '''
+        ).! ~ "'" ~ ")" ~
         StringValParser.variable ~
-        "\"" ~
-        CharsWhile(_ != '\"').! ~
-        "\"" ~
         ")"
     )
-      .map(p => Aggregate.GROUP_CONCAT(p._1, p._2))
+      .map(p => Aggregate.GROUP_CONCAT(p._2, p._1))
```
TomyMeren, line 36: why this change?

TomyMeren (approved):
LGTM!

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/AggregateParser.scala
patofp: I don't know exactly because this was part of the code that @JuanMGN did, but for what I can see the Sparql Algebra switch places between the variable and the separator respect to our case class.

E.g:
```
(group_concat (separator '|') ?v)
```

patofp (approved):

PR #531, opened 2021-08-31
introduce Relational typeclass and typed dataframes

pepegar:
# Description

## `Relational` 

it's a typeclass that describes the required relational algebra operations needed to compile SparQL expressions.  So far we've an instance for untyped dataframes `DataFrame @@ Untyped`, but we'll introduce one for typed dataframes too.

# Type of change

- `bug` if it fixes a bug

CLOSES AIPL-3924


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -349,46 +393,46 @@ object Engine {
...
         .map(columns => df.agg(columns.head, columns.tail: _*))
   }
 
   private def evaluateOrder(
       conds: NonEmptyList[ConditionOrder],
-      r: Multiset
-  ): M[Multiset] = {
-    M.ask[Result, Config, Log, DataFrame].flatMapF { config =>
+      r: Multiset[DataFrame @@ Untyped]
+  )(implicit sc: SQLContext): M[Multiset[DataFrame @@ Untyped]] = {
```
patofp, line 406: I can see that some of these methods now receive the implicit `SQLContext` but I can't see the reason.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -36,8 +41,10 @@ import java.{util => ju}
 
 object Engine {
 
-  def evaluateAlgebraM(implicit sc: SQLContext): AlgebraM[M, DAG, Multiset] =
-    AlgebraM[M, DAG, Multiset] {
+  def evaluateAlgebraM(implicit
+      sc: SQLContext
+  ): AlgebraM[M, DAG, Multiset[DataFrame @@ Untyped]] =
```
patofp, line 46: I wonder if it make sense to you and also if it would work to use an alias for `Multiset[DataFrame @@ Untyped]` or maybe for `Dataframe @@ Untyped` or both? I think it could save some boilerplate, what do you think?

modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
```diff
@@ -0,0 +1,319 @@
...
+  *
+  * =Laws=
+  *
+  * {{{
+  * Relational[A].union(Relational[A].empty, Relational[A].empty) <=> Relational[A].empty
+  * Relational[A].union(df, Relational[A].empty) <=> df
+  * Relational[A].union(Relational[A].empty, df) <=> df
+  * }}}
+  */
+@typeclass trait Relational[A] {
```
patofp, line 24: Nice job @pepegar!

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -286,36 +320,46 @@ object Engine {
...
 
   private def evaluateGroup(
       vars: List[VARIABLE],
       func: List[(VARIABLE, Expression)],
-      r: Multiset
-  ): M[Multiset] = {
-    val df = r.dataframe
+      r: Multiset[DataFrame @@ Untyped]
+  )(implicit sc: SQLContext): M[Multiset[DataFrame @@ Untyped]] = {
+    val groupedDF: RelationalGroupedDataset @@ Untyped =
```
patofp, line 340: Hmm what is this new `RelationalGroupedDataset`?

modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
```diff
@@ -0,0 +1,319 @@
...
+
+      def collect(df: DataFrame @@ Untyped): Array[Row] =
+        df.unwrap.collect()
+
+      def toDataFrame(df: DataFrame @@ Untyped): DataFrame =
+        df.unwrap
+    }
+}
+
+/** types of the RelationalGrouped typeclass can be grouped in a relational fashion.
```
patofp, line 263: I don't quite understand in what cases we may want this behaviour or for example when we could have another instance of this typeclass.

patofp (commented):
Awesome job @pepegar! So the next step would be to have an instance of `Relational` for the `Dataframe @@ Typed` and implement all the algebra for that instance right?

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
pepegar: In this case, I prefer having the types as they're instead of creating a type alias.

modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
pepegar: Oh, yeah, we really need that to be able to perform grouping operations in Spark.

When you do a df.groupBy in Spark, it does return RelationalGroupedDataframe type, not DataFrame, with its own set of operations.  We need this to separate these operations from DataFrame's.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
pepegar: good catch.  Initially I needed SqlContext for the Relational instance, but not anymore :)

patofp (approved):

PR #533, opened 2021-09-02
Typed dataframes: add DataFrame typer

pepegar:
# Description

This PR introduces the `DataFrameTyper` layer to the engine.  This PR also changes the RdfFormatter to make it able to reformat typed dataframes too.

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/Compiler.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/Compiler.scala
@@ -24,11 +24,12 @@ object Compiler {
 
   // scalastyle:off
   def compile(df: DataFrame, query: String, config: Config)(implicit
-      sc: SQLContext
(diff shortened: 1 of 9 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/Compiler.scala
```diff
@@ -50,19 +50,59 @@ object Compiler {
...
+    * +---v---------+
+    * |rdfFormatter |
+    * +-------------+
+    * }}}
+    *
+    * @param sc Spark's [[SQLContext]], required for the engine to run
     * @return
     */
-  private def compiler(df: DataFrame)(implicit
+  private def compiler[T: Basis[DAG, *]](implicit
```
patofp, line 94: Nice!

patofp (approved):
LGTM!

PR #534, opened 2021-09-03
Add bnode function

TomyMeren:
# Description

Implementation of the BNODE function
https://www.w3.org/TR/sparql11-query/#func-bnode

## Type of change

enhacement


patofp (changes requested):

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
```diff
@@ -222,4 +223,32 @@ object FuncTerms {
...
+      if (str.isEmpty) {
+        prefix + java.util.UUID.randomUUID().toString
+      } else {
+        val strToArrayOfBits = (str + rowID.toString).toArray.map(_.toByte)
+        rowID = rowID + 1
+        prefix + java.util.UUID.nameUUIDFromBytes(strToArrayOfBits).toString
+      }
+    }
+
+    val udfFoo = udf(nameToArrayOfBits)
```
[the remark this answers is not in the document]
TomyMeren: I have not excuse 😅

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
```diff
@@ -222,4 +223,32 @@ object FuncTerms {
     val endPos   = length(u)
     u.substr(startPos, endPos)
   }
+
+  /** @param col
```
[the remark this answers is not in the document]
TomyMeren: I have not excuse again😅

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
```diff
@@ -222,4 +223,32 @@ object FuncTerms {
...
+    * @return
+    */
+  def bNode(col: Column): Column = {
+    val prefix = "_:"
+
+    // Necesary var to generate diferents UUIDS between rows names when you specific de label name
```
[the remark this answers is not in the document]
TomyMeren: The think is that we need to generates differents UUID per row. 

When you don't specific a input parameter, the function use UUID.randomUUID(). and everything is ok.

When you specific a Input parameter we use UUID.nameUUIDFromBytes(value:Array[Bytes]) you can specific a array of bytes(witch we get from the string input parameter) and generates the same UUID if you specific the same name. So far so good. The problem is that you need one specific and different UUID per row and you need replicate the same UUIDs per row for the same name.

I think that if we forget the pure functional part It isn't a bad solutions. I have doubts with its behaviour in a distributed cluster. But I per node generates his own part of BNODEs all will be fine.

In any case I will try to change to a specific alternative 😉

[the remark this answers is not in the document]
pepegar: I think we need to rethink this.

The way I understand the [spec](https://www.w3.org/TR/sparql11-query/#func-bnode), we should generate:

1. random values for each time the `bnode()` function is called without params.  That part we're doing it in the `if` statement. ✅ 
2. a random value for each different literal passed to the `bnode(literal)` function.  What you're doing here is adding an extra string to the `literal` input of the function, effectively changing the value of the result.  I think we shouldn't do that, but instead leave the input as input to `nameUUIDFromBytes`, right?

If you want we can discuss it further.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/BNodeSpec.scala
```diff
@@ -0,0 +1,333 @@
...
+        val countRow = count(col(Evaluation.renamedColumn)).over(wind)
+        val countDistinctRow = org.apache.spark.sql.functions
+          .size(collect_set(Evaluation.renamedColumn).over(wind))
+
+        Evaluation.eval(
+          df,
+          Some(countRow.equalTo(countDistinctRow)),
+          query,
+          expected
+        )
```
pepegar, line 202: In this test, I'd also test that all values of the resulting column are equal

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/BuiltInFuncParser.scala
```diff
@@ -181,6 +182,14 @@ object BuiltInFuncParser {
     P("(" ~ strUuid ~ ")")
       .map(f => STRUUID())
 
+  def bNodeParen[_: P]: P[BNODE] =
+    P("(" ~ bNode ~ ")")
+      .map(f => BNODE())
+
+  def bNodeWithNameParen[_: P]: P[BNODE] =
+    P("(" ~ bNode ~ ExpressionParser.parser ~ ")")
+      .map(f => BNODE(f))
```
pepegar, line 191: 👍🏻

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
```diff
@@ -87,6 +87,9 @@ object BuiltInFunc {
   )                          extends BuiltInFunc
   final case class UUID()    extends BuiltInFunc
   final case class STRUUID() extends BuiltInFunc
+  final case class BNODE(
+      s: Expression = StringVal.STRING("")
```
pepegar, line 91: how do you feel about changing the type of this to `Option[Expression]`?  I think it will be easier to compare stuff with `None` down the line than comparing it to an empty string.

Also, if the caller actually passes a `""`, since it's a valid literal, all generated UUIDs should be equal.

pepegar (changes requested):
Great job @TomyMeren , here are a couple of comments I think we should review.

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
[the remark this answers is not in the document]
TomyMeren: That is like I did my first version of the function. This solutions looked fin, with the same tab name I can build de same Blank node name. 
The problem is that If you do that all the rows have the same value. I saw a different behaviour in the sprawl end point.


Please try this Query

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/BNodeSpec.scala
TomyMeren: In which cases could  the results be equals?

If you speak about the case when create two blank nodes with the same input name, that test are in the line 270. 🙂

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
TomyMeren: Yeah It is a good Idea. I use the "" because that is how It has been used In the FuncStrings.replace function.

Is the parser ready for Option input?

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
```diff
@@ -222,4 +229,44 @@ object FuncTerms {
...
+    val udfSpecificUUIDBNodeName = udf(specificUUIDBNodeName)
+
+    maybeCol match {
+      case None => udfRandomUUIDBNodeName()
+      case Some(col: Column) =>
+        udfSpecificUUIDBNodeName(extractStringLiteral(col))
+    }
+  }
+
+//  def bNode(maybeString: Option[String]): Column =
```
patofp, line 267: This code should be deleted.

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
```diff
@@ -214,6 +214,11 @@ object FindVariablesOnExpression {
         case SECONDS(s)                      => s
         case TIMEZONE(s)                     => s
         case TZ(s)                           => s
+        case BNODE(s) =>
```
patofp, line 217: Just like a note, when you are going to retrieve what is inside the Some() or something else, it is normal to use:
`s.getOrElse(Set.empty[VARIABE])`, instead of the pattern matching.

modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
```diff
@@ -241,6 +241,11 @@ object ToTree extends LowPriorityToTreeInstances0 {
           case ExpressionF.SECONDS(e)  => Node(s"SECONDS", Stream(e))
           case ExpressionF.TIMEZONE(e) => Node(s"TIMEZONE", Stream(e))
           case ExpressionF.TZ(e)       => Node(s"TZ", Stream(e))
+          case ExpressionF.BNODE(s) =>
+            s match {
```
patofp, line 245: Also in this case when there are simple computations on the Some and on the None cases, you can do:

```
s.map(si => Node("BNODE", Stream(si)).getOrElse(Leaf(BNODE"))
```

It is equivalent to the pattern matching you are doing, but it is up to you.
IMHO I try to avoid pattern matching whenever it is possible and if it doesn't affect readability.

modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
```diff
@@ -290,6 +290,11 @@ object QueryExtractor {
       case SECONDS(e)                      => s"(seconds $e)"
       case TIMEZONE(e)                     => s"(timezone $e)"
       case TZ(e)                           => s"(tz $e)"
+      case BNODE(s) =>
+        s match {
```
patofp, line 294: Same here, you can do:

`s.map(si => s"(bnode $si")).getOrElse("(bnode)")`

modules/engine/src/main/scala/com/gsk/kg/engine/functions/FuncTerms.scala
TomyMeren: Done!

modules/engine/src/main/scala/com/gsk/kg/engine/analyzer/FindUnboundVariables.scala
TomyMeren: Yeah I knew that :sweat_smile: .

modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
TomyMeren: Done!

modules/engine/src/main/scala/com/gsk/kg/engine/QueryExtractor.scala
TomyMeren: Done!

PR #540, opened 2021-09-07
Property path reverse final

patofp:
# Description

This PR adds support for reverse property path. E.g:

```

SELECT ?s ?o
WHERE {
 ?s ^(foaf:knows/foaf:name/foaf:mbox/foaf:surname) ?o .
}
```

The way this property path is implemented varies from the rest of the property path by rewriting the DAG of the query. This rewrite has the next phases for the example where we have a `Path` that is the unit of evaluation of a property path expression:

In this example we have the part of the DAG that will be rewrite that it has been generated from the parser:

```
Path(
  VARIABLE(s),
  Reverse(
    SeqExpression(
      SeqExpression(
        SeqExpression(
          Uri(<http://xmlns.org/foaf/0.1/knows>),
          Uri(<http://xmlns.org/foaf/0.1/name>)
        ),
        Uri(<http://xmlns.org/foaf/0.1/mbox>)
      ),
      Uri(<http://xmlns.org/foaf/0.1/surname>)
    )
  ),
  VARIABLE(?o),
  List(GRAPH_VARIABLE),
  false
)

```

1. In first place we remove the upper `Path` expression and all `Reverse(pe: PropertyExpression)` are pushed down into all the nested Property Expressions until it finds a `Uri(s: String)` PropertyExpression. This is done by a `futumorphism` (top-down) that allows to accumulate evaluations done before pushing down the `Reverse` recursively.

```
SeqExpression(
  SeqExpression(
    SeqExpression(
      Reverse(
        Uri(<http://xmlns.org/foaf/0.1/knows>)
      ),
      Reverse(
        Uri(<http://xmlns.org/foaf/0.1/name>)
      )
    ),
    Reverse(
      Uri(<http://xmlns.org/foaf/0.1/mbox>)
    )
  ),
  Reverse(
    Uri(<http://xmlns.org/foaf/0.1/surname>)
  )
)
```

2. Once we have all `Reverse` property expression pushed down in the next phase we do some replacements:
- For every `Uri` expression we replace with a `Path` expression with `reverse` field set to `false`, or if we found a `Reverse(Uri)` we replace it with a `Path` with `reverse` field set to `true`.
- For every `SeqExpression` we replace for a `Join`.
- For every `Alternative` we replace with a `Union`.
This phase is done with a `anamorphism` (top-down) because we need to match `Reverse(Uri)` before `Uri`.

```
Join(
  Join(
    Join(
      Path(
        VARIABLE(?s),
        Uri(<http://xmlns.org/foaf/0.1/knows>),
        VARIABLE(?o),
        List(GRAPH_VARIABLE),
        true
      ),
      Path(
        VARIABLE(?s),
        Uri(<http://xmlns.org/foaf/0.1/name>),
        VARIABLE(?o),
        List(GRAPH_VARIABLE),
        true
      )
    ),
    Path(
      VARIABLE(?s),
      Uri(<http://xmlns.org/foaf/0.1/mbox>),
      VARIABLE(?o),
      List(GRAPH_VARIABLE),
      true
    )
  ),
  Path(
    VARIABLE(?s),
    Uri(<http://xmlns.org/foaf/0.1/surname>),
    VARIABLE(?o),
    List(GRAPH_VARIABLE),
    true
  )
)
```

3. The last phase is only needed if we have `Join` expressions from the previous phase. This is because we now that this query, for example this query:

```

SELECT ?s ?o
WHERE {
 ?s foaf:knows/foaf:name ?o .
}
```

Is equivalent to this one:

```

SELECT ?s ?o
WHERE {
 ?s foaf:knows ?x1 .
 ?x1 foaf:name ?o .
}
```

If you take a close look you can see that there is an intermediate variable `?x1` that has to be created, so this is the idea of this phase. In case we found nested `Join` expression, we should create this variables.

This phase is done with an `histomorphism` (down-top) that will allow to accumulate the result of evaluations done previously in subtrees.

So this would be the result after applying the last phase to our example:

```
Join(
  Join(
    Join(
      Path(
        VARIABLE(?s),
        Uri(<http://xmlns.org/foaf/0.1/knows>),
        VARIABLE(?_afd79d53-111a-4509-80ec-525765824ea9),
        List(GRAPH_VARIABLE),
        true
      ),
      Path(
        VARIABLE(?_afd79d53-111a-4509-80ec-525765824ea9),
        Uri(<http://xmlns.org/foaf/0.1/name>),
        VARIABLE(?_deef1717-b0ad-455f-bbdd-f7331281948a),
        List(GRAPH_VARIABLE),
        true
      )
    ),
    Path(
      VARIABLE(?_deef1717-b0ad-455f-bbdd-f7331281948a),
      Uri(<http://xmlns.org/foaf/0.1/mbox>),
      VARIABLE(?_7daea15a-b530-4f9e-86e9-257f8db0e0e4),
      List(GRAPH_VARIABLE),
      true
    )
  ),
  Path(
    VARIABLE(
      ?_7daea15a-b530-4f9e-86e9-257f8db0e0e4),
      Uri(<http://xmlns.org/foaf/0.1/surname>),
      VARIABLE(?o),
      List(GRAPH_VARIABLE),
      true
    )
  )
```

Finally we can see the chain of replacement variables.

After this the engine will now how to evaluate every `Path` expression and if it has the reverse flag, switch the direction of the triple.

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/PropertyPathRewrite.scala
```diff
@@ -0,0 +1,166 @@
...
+      .pure[M]
+      .ifM(
+        Log.debug(
+          "Optimizer(ReversePathRewrite)",
+          s"resulting query: ${result.toTree.drawTree}"
+        ),
+        ().pure[M]
+      ) *>
+      result.pure[M]
+  }
```
pepegar, line 165: Can you add some test for this method that compare the output of `ToTree`?  I think it can serve as great documentation in the tests to see how this transforms the DAG & PropertyPaths

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/PropertyPathRewriteSpec.scala
```diff
@@ -0,0 +1,1477 @@
...
+import com.gsk.kg.sparqlparser.TestUtils
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
+
+class PropertyPathRewriteSpec
+    extends AnyWordSpec
+    with Matchers
+    with ScalaCheckDrivenPropertyChecks
```
pepegar, line 24: Are you using scalacheck in this test?

```suggestion
```

pepegar (commented):
Looks good to me, just added a couple of comments :)

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/PropertyPathRewrite.scala
patofp: Yeah, I will add some tests and documentation.

patofp: Done

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/PropertyPathRewriteSpec.scala
patofp: Done

pepegar (approved):

PR #546, opened 2021-09-13
fixed bug where graph variables was being set to null on union operation

patofp:
This PR fixed a bug where the graph variable column was being set to null when performing a union operation.

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
@@ -54,6 +54,7 @@ import simulacrum.typeclass
   def map[U: Encoder](df: A, fn: Row => U): A
   def flatMap[U: Encoder](df: A, fn: Row => TraversableOnce[U]): A
   def collect(df: A): Array[Row]
+  def show(df: A, truncate: Boolean): Unit
   def toDataFrame(df: A): DataFrame
 }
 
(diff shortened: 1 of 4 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
```diff
@@ -54,6 +54,7 @@ import simulacrum.typeclass
   def map[U: Encoder](df: A, fn: Row => U): A
   def flatMap[U: Encoder](df: A, fn: Row => TraversableOnce[U]): A
   def collect(df: A): Array[Row]
+  def show(df: A, truncate: Boolean): Unit
```
shihch, line 57: It's fine here but seems not related to the issue.

shihch (approved):
Looks good as far as I can tell.

modules/engine/src/main/scala/com/gsk/kg/engine/relational/relational.scala
patofp: Yes, that is true. It doesn't have anything to do with the bug but is a useful method for debugging.

Strictly it shouldn't be on this PR as it is an enhancement and it has nothing to do with the bug-fix, but I think it doesn't do any harm.

What is your opinion, should we remove it from this PR?

shihch: Please go ahead to merge. I just ask for clarification. As long as the function is used, it does not matter with which pull request.

patofp: Cool!

PR #549, opened 2021-09-15
added support for default and named graphs to property paths

patofp:
This PR adds support for default and named graphs when using property paths expressions.

## Type of change

- `bug` if it fixes a bug


shihch (approved):
LGTM

PR #566, opened 2021-09-28
Add complex property path test

patofp:
# Description

This PR adds integration tests of complex property paths, by combining some of the defined property paths.
It also fixes some minor bugs:
- Property path DAG rewrite when some property paths.
- Return duplicates triples on some corner cases.

## Type of change

- `bug` if it fixes a bug

### DIFF
--- a/modules/engine/src/main/scala/com/gsk/kg/engine/functions/PathFrame.scala
+++ b/modules/engine/src/main/scala/com/gsk/kg/engine/functions/PathFrame.scala
@@ -177,13 +177,17 @@ object PathFrame {
       val cols = Seq(SIdx.toString, PIdx.toString, oSlice.toString) :+ gCol
 
       val nPaths = pathFrame
-        .select(cols.map(col))
         .filter(cols.foldLeft(lit(true)) { case (acc, elem) =>
           acc && pathFrame.getColumn(elem).isNotNull
         })
-        .withColumnRenamed(s"$oSlice", s"$OIdx")
 
-      nPaths
+      val selection = (SIdx to oSlice).map(_.toString) :+ gCol
+
+      val distinctRows = nPaths.select(selection.map(col)).distinct
+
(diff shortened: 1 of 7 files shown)


modules/engine/src/main/scala/com/gsk/kg/engine/functions/PathFrame.scala
```diff
@@ -240,7 +244,7 @@ object PathFrame {
   ): Monoid[PathFrame] =
     new Monoid[PathFrame] {
       def combine(x: PathFrame, y: PathFrame): PathFrame =
-        merge(x, y).distinct
```
pepegar, line 243: 👌🏻

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/PropertyPathRewrite.scala
```diff
@@ -390,9 +465,103 @@ object PropertyPathRewrite {
...
+//      case OneOrMore(Alternative(pel, per)) =>
+//        Union(
+//          Coattr.pure(OneOrMore(pel)),
+//          Coattr.pure(OneOrMore(per))
+//        )
+//      case OneOrMore(SeqExpression(pel, per)) =>
+//        Union(
+//          Coattr.pure(OneOrMore(pel)),
+//          Coattr.pure(OneOrMore(per))
+//        )
```
pepegar, line 523: Do we want to leave this code commented?

pepegar (approved):
LGTM

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/PropertyPathRewrite.scala
patofp: this code can be deleted 😉

PR #573, opened 2021-10-01
Fix property path rewrite when alt seq

patofp:
# Description

This PR fixes some queries that where failing due to DAG rewrite:

```

SELECT ?s ?o
WHERE {
 ?s (foaf:knows|foaf:name)* ?o .
}
```

```

SELECT ?s ?o
WHERE {
 ?s (foaf:knows/foaf:name)* ?o .
}
```

*When fixing we found out that some of the assumptions on the semantics of the DAG rewrite where run.* So we decided to revert the rewrite and implement some of the property path in a straightforward approach. Some the changes are:

- PropertyPathRewrite has been removed.
- Reverse property path has been implemented in FuncProperty
- Sequence and Alternative property path has been reverted to the version of FuncProperty with some changes.
- NotOnOf property path implementation has been refactored.
- Uri property path implementation has been refactored.
- Now all property paths returns a DataFrame instead of a Either[Column, Dataframe]
- Added tests for some of this implementations.

## Type of change

- `bug` if it fixes a bug


pepegar (approved):
LGTM, nice work cleaning up property paths!

PR #577, opened 2021-10-05
Add bnode function (from gsk repo)

TomyMeren:
# Description

This pull request is a copy from another how start in a fork in my personal GitHub

Implementation of the BNODE function
https://www.w3.org/TR/sparql11-query/#func-bnode

## Type of change

enhacement


patofp (approved):
LGTM!

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/BNodeSpec.scala
```diff
@@ -0,0 +1,362 @@
+package com.gsk.kg.engine.compiler
+import org.apache.spark.sql.DataFrame
```
pepegar, line 2:
```suggestion
package com.gsk.kg.engine.compiler

import org.apache.spark.sql.DataFrame
```

pepegar (approved):
LGTM, just a minor quirk :D

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/BNodeSpec.scala
TomyMeren: Can it change in another PR?

PR #596, opened 2021-11-02
Extra hierarchy of sparql at Order 

TomyMeren:
# Description

We have added a top level hierarchy containing SPARQL that follows the following pattern(ascending order).

1. - (Lowest) no value assigned to the variable or expression in this solution.
2. - Blank nodes
3. - IRIs
4. - RDF literals

Within each category the normal order continues.

https://www.w3.org/TR/sparql11-query/#modOrderBy

in descending order is exactly the opposite

## Type of change

- `bug` if it fixes a bug


patofp (approved):
Nice solution, good job @TomyMeren!

pepegar (approved):
Great job @TomyMeren !

PR #600, opened 2021-11-08
Remove the duplicate in the prefix name of the blank nodes created by de BNODE() function

TomyMeren:
# Description

The prefix "_:" does't contains relevant information show no need to add in value column . It is added when want to show it. 
That is why the blank nodes generates with BNODE contain a duplicate prefix like this one:

<div>"_:_:123e4567-e89b-12d3-a456-426614174000"</div>

.

In this PR, we remove the duplicate in the prefix name of the blank nodes created by de BNODE() functions and change the test to detect it if it happens again.

## Type of change

- `bug` if it fixes a bug


patofp (approved):
LGTM!

PR #601, opened 2021-11-09
Add reasoning engine

patofp:
# Description

Adds backward-chaining inference with query rewrite.

## Type of change

- `bug` if it fixes a bug


modules/parser/src/main/scala/com/gsk/kg/config/Config.scala
```diff
@@ -5,9 +5,10 @@ final case class Config(
...
     formatRdfOutput: Boolean,
     typeDataframe: Boolean,
-    reasoningInference: Boolean
+    forwardInference: Boolean,
+    backwardInference: Boolean
 )
 
 object Config {
-  val default: Config = Config(true, false, false, true, false)
+  val default: Config = Config(true, false, false, true, false, false)
```
puremachinery, line 13: Curious what happens when both config settings are true (forward & backward). Sbt won't build this PR for me so I can't test it directly right now.

puremachinery (approved):
LGTM

modules/parser/src/main/scala/com/gsk/kg/config/Config.scala
patofp: The short answer will be 💣  Hahahah

Yeah, they should be mutually exclusive, I'll change the configuration so it is an enumeration. Something like:
inference-mode: {forward | backward | none}

modules/engine/src/main/scala/com/gsk/kg/engine/Compiler.scala
```diff
@@ -13,39 +13,46 @@ import com.gsk.kg.Graphs
...
 import com.gsk.kg.sparqlparser.Result
 
 object Compiler {
 
+  type KG     = DataFrame
+  type Schema = DataFrame
+
   // scalastyle:off
   def compile(
       kg: DataFrame,
```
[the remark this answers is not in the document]
patofp: The motivation of this type aliases is to improve readability on the compiler method. You can imagine if we substitute KG and Schema for DataFrame in the method bellow it will be harder than it already is to follow the flow:

```
private def compiler[T: Basis[DAG, *]](implicit
      sc: SQLContext
  ): Phase[(String, KG, Schema), DataFrame] = {

    val splitSchema: Phase[
      (String, KG, Schema),
      ((String, Schema), (KG, Schema))
    ] =
      Kleisli[M, (String, KG, Schema), ((String, Schema), (KG, Schema))] {
        case (query, kg, schema) =>
          ((query, schema), (kg, schema)).pure[M]
      }

    val queryFlow: Phase[(String, Schema), T] = {
      val dagAndGraphs = parser >>>
        transformToGraph.first[Graphs] >>>
        staticAnalysis.first[Graphs] >>>
        optimizer

      val dagAndGraphsMergedWithSchema: Phase[
        (String, Schema),
        (T, Schema)
      ] = dagAndGraphs.first[Schema]

      dagAndGraphsMergedWithSchema >>> backwardChainingInference >>> SubqueryPushdown.phase
    }

    val dataFlow: Phase[(KG, Schema), DataFrame] =
      forwardChainingInference >>> dataFrameTyper

    splitSchema >>> (queryFlow *** dataFlow) >>> engine >>> rdfFormatter
  }
```

It came up to me the idea of propagating to the rest of the methods, specially to the `compile` method that you pointed out. But I decided not to do that because that method act as the API and will be called from outside the engine thus I think is better to keep them as DataFrame that is something that developers that use the API are already familiar with.

modules/engine/src/main/scala/com/gsk/kg/engine/inference/BackwardChainingInference.scala
```diff
@@ -0,0 +1,102 @@
...
+            bgpR(quads)
+          case first :: Nil =>
+            first
+          case first :: second :: Nil =>
+            joinR(first, second)
+          case first :: second :: tail =>
+            tail.foldLeft(joinR(first, second)) { case (acc, elem) =>
+              joinR(acc, elem)
+            }
+        }
```
TomyMeren, line 75: A silly question: Is the specification of line 69 and 70 necessary?

TomyMeren (approved):
LGMT!

modules/engine/src/main/scala/com/gsk/kg/engine/inference/BackwardChainingInference.scala
TomyMeren: Sorry I have clicked below 😆

PR #602, opened 2021-11-11
Add support property path inverse reverse

TomyMeren:
# Description

Fix the corner case in the property path when we mix inverse (!) and reverse(^) in in this order

SELECT ?s ?o
WHERE { ?s !(^foaf:knows) ?o . }

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/PropertyPathsSpec.scala
```diff
@@ -9413,7 +9413,7 @@ class PropertyPathsSpec
         }
 
         // TODO: Fix this one
```
patofp, line 9415: You can remove the TODO

patofp (approved):

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/PropertyPathsSpec.scala
TomyMeren: true!

PR #609, opened 2021-11-23
Add comparation sparql hierarchy

TomyMeren:
# Description

Change the comparison functions (equal, gt,lt, gte and lte) to take in account the behaviour between nulls, Uri, blank nodes numbers and literals. 

Basically, in the equals case for Uri and blank nodes, the value is compared as if it were a literal.
In all other cases (gt,lt, gte and lte) they are not compare and the result is always null. 

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -96,6 +100,55 @@ object Engine {
...
+    quad.getPredicates
+      .groupBy(_._2)
+      .map { case (_, vs) =>
+        vs.map { case (pred, position) =>
+          val col = relational.getColumn(position)
+          col === DataFrameTyper.parse(lit(pred.s))
+        }.foldLeft(lit(false))(_ || _)
+      }
+      .foldLeft(lit(true))(_ && _)
+  }
```
patofp, line 129: This is just more of a good practice. To make it easy to review the pull request try to avoid moving methods around the file if there is no need for. There could be exceptions to this like refactors.

I know that this method existed previously but I can't easily see if there have been changes. 😉

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -96,6 +100,55 @@ object Engine {
...
+
+    new GenericRowWithSchema(
+      Array(
+        value,
+        tpe,
+        null // scalastyle:off
+      ),
+      typedField
+    )
+  }
```
patofp, line 150: Same for this one.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -736,24 +754,9 @@ object Engine {
...
-        null // scalastyle:off
-      ),
-      typedField
-    )
-  }
+  final case class OrderRank(
+      sparqlHierarchy: Int,
+      stringOrder: String,
+      numberOrder: Double
+  )
```
patofp, line 761: Same for this one.

patofp (approved):
LGTM! Please take a look at the comments tho.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -47,6 +47,10 @@ object Engine {
       StructField("lang", StringType, true)
     )
   )
+  private val createDescribeTriple: (StringVal, Int) => Quad = { (subject, i) =>
+    def createVar(letter: String): VARIABLE = VARIABLE(s"?$letter$i")
+    Quad(subject, createVar("p"), createVar("o"), Nil)
+  }
```
patofp, line 53: Same here.

patofp (approved):

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
TomyMeren: Yes @patofp you are absolutely right. I can't understand why the methods are moved. I don't know if it happened by merging the master changes in that branch or moving it manually.

PR #616, opened 2021-12-02
added backward chaining for property path triples (single triple)

patofp:
# Description

Added support for property path triples on backward chaining (unique triple)

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/inference/BackwardChainingInference.scala
```diff
@@ -11,86 +12,174 @@ import com.gsk.kg.config.Config
...
+          unionR(acc, elem)
+        })
+    }
+
+    // Generate sub-query
+    maybeWhereClause.map { whereClause =>
+      projectR(List(StringVal.VARIABLE(p.s.s)), distinctR(whereClause))
+    }
+  }
+
```
TomyMeren, line 113: Take in account that backwardPath and backwardQuad  are really similar, would it be better unify the two methods into one and include a flag to differentiate them?

TomyMeren (approved):
LGTM!

patofp:
Closing this PR due to changes done on BGP. https://github.com/gsk-aiops/bellman/pull/624

PR #619, opened 2021-12-07
Add undef to values

TomyMeren:
# Description

Queries with VALUES statement can accept UNDEF on the inline data. E.g:

```
PREFIX dc:   <http://purl.org/dc/elements/1.1/> 
PREFIX :     <http://example.org/book/> 
PREFIX ns:   <http://example.org/ns#> 

SELECT ?book ?title ?price
{
   ?book dc:title ?title ;
         ns:price ?price .
   VALUES (?book ?title)
   { (UNDEF "SPARQL Tutorial")
     (:book2 UNDEF)
   }
}
```

The algebra generated for this query is:

```
(project (?book ?title ?price)
  (join
    (bgp
      (triple ?book <http://purl.org/dc/elements/1.1/title> ?title)
      (triple ?book <http://example.org/ns#price> ?price)
    )
    (table (vars ?book ?title)
      (row [?title "SPARQL Tutorial"])
      (row [?book <http://example.org/book/book2>])
    )))
```
The problem with this and similar queries is that some rows in the table part don't have all the vars values defined. For example in this Algebra the first rows don't have "?book" and the second one don't have "?title" defined.

That generates "nulls" values in the join, the normal join doesn't cross the nulls values and we lose it.

To solve this, we create a new type of join, the "tablejoin" that crosses nulls values. To apply it, we first detect this situation in the optimisation part and change the "join" to "tablejoin" in the algebra part.


patofp (changes requested):
I think there is no need to introduce the DAG.TableJoin, I don't think is part of the Sparql Algebra, and also a new Optimizer phase. You can do this check:
```
val lengthOfsmallerRow: Int = rows
            .map(row => row.tuples.length)
            .min
```
inside the Engine. evaluateJoin.

What are your thoughts on this. We can discuss it. 😉

TomyMeren:
A lot of new changes to do in the same PR. I am closing this one and I will create a new Branch with a new version of this enhancement.

PR #624, opened 2021-12-16
Property path and quads to sequence

TomyMeren:
# Description

Create the DAG.quad and Adapt the DAG.BGP to accept al the recursive type.
With that changes we seek to bring together in the same structure the DAG.Quad and the DAG.Path, and them apply after
a reordering optimisation between their variables.

Create the DAG.quad and Adapt the DAG.BGP to accept al the recursive type.
With that changes we seek to bring together in the same structure the DAG.Quad and the DAG.Path, and them apply after
a reordering optimisation between their variables.

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/GraphsPushdownSpec.scala
```diff
@@ -24,12 +23,16 @@ class GraphsPushdownSpec
...
-    list.map(assert)
+  val assertForAllQuads: List[Any] => (
+      DAG.Quad[T] => Assertion
+  ) => Unit = { list: List[Any] => assert: (DAG.Quad[T] => Assertion) =>
+    list.map { case DAG.Quad(s, p, o, g) =>
+      assert(DAG.Quad(s, p, o, g))
+    }
   }
 
+  // TODO(tomy): Create test for paths as well?
```
patofp, line 34: I suppose this can be done when you continue with the BGP optimization task right?

modules/engine/src/test/scala/com/gsk/kg/engine/scalacheck/DAGArbitraries.scala
```diff
@@ -89,8 +80,24 @@ trait DAGArbitraries extends CommonGenerators with ExpressionArbitraries {
...
+      stringValGenerator,
+      propertyExpresionGenerator,
+      stringValGenerator,
+      stringValGenerator
+    ).mapN((s, p, o, g) => DAG.Path(s, p, o, g :: Nil))
   def bgpGenerator[A](implicit A: Arbitrary[A]): Gen[BGP[A]] =
-    Gen.listOf(quadGenerator).map(BGP(_))
+    Gen
+      .listOf(Gen.oneOf(quadGenerator, pathGenerator))
+      .map(x => BGP(x.map(_.asInstanceOf[A])))
```
patofp, line 100: Is this casting safe?

modules/parser/src/test/scala/com/gsk/kg/sparqlparser/ShortQueryTestSpec.scala
```diff
@@ -9,6 +9,11 @@ import org.scalatest.flatspec.AnyFlatSpec
 
 class ShortQueryTestSpec extends AnyFlatSpec with TestUtils with TestConfig {
 
+  implicit val exprToQuad = (expr: Expr) =>
+    expr match {
+      case Expr.Quad(s, p, o, g) => Expr.Quad(s, p, o, g)
+    }
+
```
patofp, line 16: I see this value duplicated across multiple tests files, can we move it to a utility, maybe TestUtils?

patofp (commented):
This file should be removed for this PR: modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/TableJoinInValuesSpec.scala

modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
```diff
@@ -35,18 +35,18 @@ object ToTree extends LowPriorityToTreeInstances0 {
...
+//      TreeRep.Node(
+//        s"Quad",
+//        Stream(
+//          t.s.s.toTree,
+//          t.p.s.toTree,
+//          t.o.s.toTree,
+//          t.g.toString().toTree
+//        )
+//      )
+//  }
```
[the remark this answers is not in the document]
TomyMeren: After including and using DAG.Quads instead of Expr.Quads in the DAG I guess we won't use it any more.
Even so I've printed some trees with Quads and it looks good, also I think some of the test would Fail.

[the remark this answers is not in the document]
TomyMeren: So I going to removed If that's Ok for you

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/GraphsPushdownSpec.scala
TomyMeren: That is! I have written two TODO's about this subject of take in account only the Quads in some match case. In the optimisation part I guess I will have to take paths into account  as well and I will create new tests with it.

modules/engine/src/test/scala/com/gsk/kg/engine/scalacheck/DAGArbitraries.scala
TomyMeren: The problem is this:

found   : List[Product with Serializable with com.gsk.kg.engine.DAG[Nothing]]
required: List[A]
     .map(BGP(_))

That asInstanceOf works but I don't now if it may be dangerous ... 

Do you know of another option to pass from DAG[A] to A?

modules/parser/src/test/scala/com/gsk/kg/sparqlparser/ShortQueryTestSpec.scala
TomyMeren: Good idea. Done!

modules/engine/src/test/scala/com/gsk/kg/engine/scalacheck/DAGArbitraries.scala
patofp: I think that the problem is that Nothing

`Gen.listOf(Gen.oneOf(quadGenerator[A], pathGenerator[A])).map(BGP(_))`

> Do you know of another option to pass from DAG[A] to A?

You don't need to do this, BGP should accept a DAG[A], but because you throw this question, I will throw you another one. Does this looks familiar to you?  F[A] => A

patofp: You can do this, define this Arbitrary implicits:

```
implicit def arbitraryQuad[T: Arbitrary]: Arbitrary[DAG.Quad[T]] =
    Arbitrary(quadGenerator[T])
  implicit def arbitraryPath[T: Arbitrary]: Arbitrary[DAG.Path[T]] =
    Arbitrary(pathGenerator[T])
```
And then use A.arbitrary:

```
def bgpGenerator[A](implicit A: Arbitrary[A]): Gen[BGP[A]] =
    Gen.nonEmptyListOf(A.arbitrary).map(_.toList).map(BGP(_))
```

The problem with this, A.arbitrary will generate an arbitrary for every implicit defined in scope. So it will not only generate Quads or Paths, there must be a way to constraint.

TomyMeren: Change Done!

patofp (approved):
LGTM

modules/engine/src/test/scala/com/gsk/kg/engine/scalacheck/DAGArbitraries.scala
TomyMeren: Problems @patofp the OpticalLawTests don't end with that new change.
I guess it is because A.arbitrary is Open to introduce all the recursive types and not just Paths and Quads.

PR #632, opened 2021-12-28
added support for property expression triples on backward chaining inference

patofp:
# Description

Adapted backward chaining to take into account property expression triples and the new BGP.

## Type of change

- `bug` if it fixes a bug


TomyMeren (approved):
LGTM!

PR #634, opened 2021-12-30
Fix nulls in join

TomyMeren:
# Description

It takes into account when we have null values on the left, right or both sides of the Join.

SparqQL always tries to get a value from a join hand instead of a null. This approach includes a possible combination of the fields

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ConstructSpec.scala
```diff
@@ -144,17 +144,23 @@ class ConstructSpec
       }
       """
 
-      Compiler
-        .compile(df, None, query, config)
-        .right
-        .get
-        .collect()
-        .toSet shouldEqual Set(
+      val result = Compiler.compile(df, None, query, config)
```
TomyMeren, line 147: @patofp can you help me check this result? I have a strange result in Jena

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/GraphSpec.scala
```diff
@@ -1829,8 +1829,18 @@ class GraphSpec
         val result = Compiler.compile(df, None, query, config)
 
         result shouldBe a[Right[_, _]]
-        result.right.get.collect.length shouldEqual 1
+        result.right.get.collect.length shouldEqual 3
         result.right.get.collect.toSet shouldEqual Set(
+          Row(
```
TomyMeren, line 1834: this is the test with which we detected the problem

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ConstructSpec.scala
patofp: I've checked the result on Jena and the result matches the previous version:

```
Set(
        Row(
          "_:test",
          "<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>",
          "<http://id.gsk.com/dm/1.0/Document>"
        ),
        Row(
          "_:test",
          "<http://id.gsk.com/dm/1.0/docSource>",
          "\"source\""
        )
```

patofp (changes requested):

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ConstructSpec.scala
TomyMeren: Resolved! Thank you @patofp.
It turns out that when a row have all the columns of the join condition null, they never cross.
I solved with a filter to remove those rows before the join.

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -370,6 +374,72 @@ final case class Multiset(
       .drop(leftGraphCol)
       .drop(rightGraphCol)
   }
+
+  def innerJoinEmptyValues(
```
patofp, line 378: I suggest to document this method, what it does and what are the steps. What do you think?

TomyMeren: Yes, sounds good!

TomyMeren: Done!

modules/engine/src/main/scala/com/gsk/kg/engine/Multiset.scala
```diff
@@ -370,6 +374,113 @@ final case class Multiset(
       .drop(leftGraphCol)
       .drop(rightGraphCol)
   }
+
+  /** Check if the df's have nulls values in any of his rows. if don't have, we
```
patofp, line 378: Nice!

patofp (approved):
LGTM! GJ!

PR #637, opened 2022-01-10
added domain and range to forward chaining

patofp:
Adds domain and range entailments support, rdfs2 and rdfs3 to forward chaining inference.

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/inference/ForwardChainingInference.scala
```diff
@@ -29,7 +29,23 @@ object ForwardChainingInference {
...
+          schema,
+          config.copy(formatRdfOutput = true)
+        )
+        rdfs3Triples <- rdfs3QueryResult(
+          kg,
+          schema,
+          config.copy(formatRdfOutput = true)
+        )
+      } yield kg union (
+        schema union
```
puremachinery, line 43: Can `schema` end up with duplicate triples as the entailments? How? Or if it can't, should it be moved out of this `distinct` scope?

modules/engine/src/main/scala/com/gsk/kg/engine/inference/ForwardChainingInference.scala
```diff
@@ -29,7 +29,23 @@ object ForwardChainingInference {
...
+          rdfs9Triples union
+          rdfs11Triples union
```
puremachinery, line 45: I see how rdfs 2, 3, and 9 can have duplicates with one another, but not how rdfs11 entailments can duplicate 2, 3, or 9. Should it be in this scope for `distinct`?

modules/engine/src/main/scala/com/gsk/kg/engine/inference/RDFSEntailments.scala
```diff
@@ -115,4 +115,68 @@ object RDFSEntailments {
         resultDf.withColumn(graphCol, emptyLit)
       }
   }
+
+  def rdfs2QueryResult(
```
puremachinery, line 119: `rdfs2QueryResult` and `rdfs3QueryResult` are nearly identical. I think the only different is in the query string. Perhaps these should be refactored into one method, which takes a query string parameter? (Or an entailment type parameter, which is elsewhere mapped to query strings).

modules/engine/src/main/scala/com/gsk/kg/engine/inference/ForwardChainingInference.scala
patofp: Yeah, thats a really good point @puremachinery. Only between rdfs2, 3 and 9 (also the `schema` can be excluded from the distinct scope) where we can have duplicates as they share predicate on the right hand side of the entailments.

modules/engine/src/main/scala/com/gsk/kg/engine/inference/RDFSEntailments.scala
patofp: Totally agree! We can abstract this kind of entailments that uses the KG to materialize triples into a method with the query string as a parameter. 👍

puremachinery (approved):
lgtm

TomyMeren (approved):
LGTM!

PR #645, opened 2022-02-04
Bug is literal non literal

TomyMeren:
# Description

IsLiteral() does not work correctly, it return Uri in the col ?o of the next Query.

df.sparql("""
SELECT ?s ?p ?o 
FROM <https://www.ncbi.nlm.nih.gov/clinvar/> 
WHERE { 
  ?s ?p ?o 
  FILTER isLiteral(?o) }
  LIMIT 10""").show(10,false) 

The correct behaviour is return True when a String is a literal and False when is a URi. For the other types return always True except for Blank Nodes which return False.

## Type of change

- `bug` if it fixes a bug


patofp (approved):
LGTM!

PR #651, opened 2022-02-15
added query entailment analyzer phase

patofp:
# Description

Added query entailment inference phase and join query and schema entailments phase.

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/inference/IntersectEntailmentAnalyzer.scala
```diff
@@ -0,0 +1,11 @@
+package com.gsk.kg.engine.inference
+
+import com.gsk.kg.config.InferenceEntailment
+
+object IntersectEntailmentAnalyzer {
+
```
TomyMeren, line 6: Is it necessary to create this object only for this sorting method?

TomyMeren (approved):
LGTM!

modules/engine/src/main/scala/com/gsk/kg/engine/inference/IntersectEntailmentAnalyzer.scala
patofp: Not at all! But as it is a Phase on it's own I decided to create it as an object instance as the rest of the Phases.
Captura de pantalla 2022-02-16 a las 16 49 19

TomyMeren: Ok, It gives more clarity to the whole process 😄

PR #652, opened 2022-02-17
Migration to Spark 3

L-Lavigne:
Cross-building with Scala 2.11 + Spark 2.4.4 and Scala 2.12 + Spark 3.1.2.


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/OrderBySpec.scala
```diff
@@ -647,6 +647,7 @@ class OrderBySpec
               "<http://example.org/Tomy>",
               "\"1.12\"^^<http://www.w3.org/2001/XMLSchema#decimal>"
             ),
+            // With Spark 3, the next two Rows must be swapped for the test to pass. TODO: investigate.
```
patofp, line 650: A revisit to the `evaluateOrder` method on `Engine:405` may be required to check whether it totally meets the [spec](https://www.w3.org/TR/rdf-sparql-query/#modOrderBy)

patofp (commented):
Just leaving here some notes for further steps we may want to look into after upgrading to Scala 2.12:

- We are using a Sparql parsing library [jena-arq 3.17.0](https://mvnrepository.com/artifact/org.apache.jena/jena-arq/3.17.0), greater versions `4.x.x` finished support for Scala 2.11, so we may what to upgrade this library also.

- Another thing is that If we are able to make SBT `crossVersion` work for Scala 2.11 and Scala 2.12 it would be relatively easy to adapt the CI pipelines to generate artifacts for both versions `.github/workflows`. 

For the rest it seems quite straightforward and that there are no big issues apart from looking into those tests for the migration. GJ!

LGTM

build.sbt
```diff
@@ -4,19 +4,18 @@ lazy val Versions = Map(
   "kind-projector"       -> "0.13.2",
   "cats"                 -> "2.0.0",
   "cats-scalacheck"      -> "0.2.0",
-  "jena"                 -> "3.17.0",
+  "jena"                 -> "4.4.0",
```
TomyMeren, line 7: Only one appointment.
Apache JENA 4.4.0 need almost java 11 to support org/apache/jena/query/Query

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/OrderBySpec.scala
TomyMeren: When The input dataset don’t contnains String, only numerics, the order is numerical, If the data have Strings the order is alphanumeric. You can test the behaviour it  in Jena. I have Input data examples if you want :)

build.sbt
L-Lavigne: Indeed, I had to switch from Java 8 to 11 in order to run it. @TomyMeren do you think this could cause a problem down the line?

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/OrderBySpec.scala
L-Lavigne: Thanks @TomyMeren , this would imply that the old behavior is the correct one and we need to fix the regression, correct?
Also if you can share your input examples and how you're processing them, that would be helpful indeed.

build.sbt
L-Lavigne: OK I see, we have Spark 2 requiring Java 8 and Jena 4 requiring Java 11. I've pushed a commit with specialized dependencies for cross-building including the JVM compilation target, let's see if it works.

build.sbt
```diff
@@ -55,8 +56,8 @@ inThisBuild(
 )
 
 lazy val buildSettings = Seq(
-  scalaVersion       := Versions("scala211"),
-  crossScalaVersions := List(Versions("scala211"), Versions("scala212")),
+  scalaVersion       := Versions("scala212"),
+  //crossScalaVersions := List(Versions("scala211"), Versions("scala212")), // Spark 3 supports 2.12 [and 2.13] only
```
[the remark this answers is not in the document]
L-Lavigne: @patofp I've set up crossbuilding dependencies in my latest commit, please take a look.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/OrderBySpec.scala
TomyMeren: Yes, the tests are correct and should pass unchanged.
You might download jena in local https://jena.apache.org/download/ , the binary distribution. and run with the next command: 

apache-jena-fuseki-4.2.0 % ./fuseki-server --update --mem  /ds

After that you can access to a local web app in http://localhost:3030/ and play with different dataset, SparkQL querys etc.

Here is an example of an input dataset

[seq_with_test_nulls_bnode_uri.ttl.zip](https://github.com/gsk-aiops/bellman/files/8131446/seq_with_test_nulls_bnode_uri.ttl.zip)

build.sbt
TomyMeren: I don't know what the consequences could be, will it be necessary that the place where the library is going to be used has java 11?

patofp (commented):
Good catch @TomyMeren with the Java version. @L-Lavigne the DataQuery API embeds the Bellman's Engine (is set as a dependency) so this could force the DataQuery API to also upgrade to Java 11. That's something to keep on mind.

Maybe we can treat the Jena 4.4.0 upgrade as a different issue or as a step after the Spark migration... 🤔 

What are your thoughts guys?

TomyMeren:
Yes in my opinion it is a different task, and I think most of the errors in the test are due to the Jena version.

build.sbt
```diff
@@ -27,6 +22,27 @@ lazy val Versions = Map(
...
+    "spark-testing-base" -> "3.2.0_1.1.1", // currently supports Spark up to 3.2.0
+  ),
+  "2.13" -> Map(
+    "jvm"                -> "11",
+    "jena"               -> "4.4.0", // requires Java 11
+    "spark"              -> "3.2.1", // see https://spark.apache.org/docs/3.2.1/index.html
+    "spark-testing-base" -> "No Scala 2.13 support yet, see https://github.com/holdenk/spark-testing-base/issues/347"
+  )
+)
+
```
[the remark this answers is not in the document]
TomyMeren: 👍

L-Lavigne:
> Yes in my opinion it is a different task, and I think most of the errors in the test are due to the Jena version.

Agreed, I'll make the Jena upgrade a different task. Unfortunately most of the test failures happen when upgrading Spark only; for those involving Jena I've indicated so in the test TODO comments.

patofp:
> > Yes in my opinion it is a different task, and I think most of the errors in the test are due to the Jena version.
> 
> Agreed, I'll make the Jena upgrade a different task. Unfortunately most of the test failures happen when upgrading Spark only; for those involving Jena I've indicated so in the test TODO comments.

Cool! Sound good @L-Lavigne!

L-Lavigne:
@patofp and @TomyMeren thanks for your help and advice on this upgrade. Please see AIPL-5763 & sub-tasks for a summary and implementation outline.

patofp:
> @patofp and @TomyMeren thanks for your help and advice on this upgrade. Please see AIPL-5763 & sub-tasks for a summary and implementation outline.

Glad to help! Taking a look to the sub-tasks, the way to go forward seems clear to me. GJ @L-Lavigne !

L-Lavigne:
@patofp & @TomyMeren Unit test failures are now fixed and PR is ready for review, please take a look. Thanks!

L-Lavigne:
Getting Scalafix errors in CI for 2.12, looking into this.

patofp (approved):
LGTM!

TomyMeren (approved):
LGTM!

L-Lavigne:
@TomyMeren @patofp I changed the Spark 3 version from 3.2 to 3.1.2 to ensure compatibility with Hudi. I also merged the latest `master` and did some refactoring to dependencies and version management in `build.sbt`. I'm hoping to merge soon in order to get a snapshot we can use in dependent projects - could you take another look? Thanks!

patofp (approved):

PR #685, opened 2022-04-22
Revert remove rows with all join columns null

patofp:
# Description

This PR reverts changes from: https://github.com/gsk-aiops/bellman/commit/f0202021e95a2ce519a5fd3a963e1fec6a8c3170
due to performance issues when performing a inner join, that triggers additional Spark Jobs

Some test that affects to null values on some operations have been ignored. Null values on join should be addressed by issue: https://myjira.gsk.com/browse/AIPL-5089 

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/GraphSpec.scala
```diff
@@ -1736,8 +1736,9 @@ class GraphSpec
         )
       }
 
+      // TODO: https://myjira.gsk.com/browse/AIPL-5089
       // TODO: This query seems not to be consistent with Jena, shall be analyzed as a case of study
-      "execute and obtain expected results when GRAPH with variable over mixed named and default graphs with JOIN" in {
+      "execute and obtain expected results when GRAPH with variable over mixed named and default graphs with JOIN" ignore {
```
L-Lavigne, line 1741: @patofp Could you clarify why this particular test fails after the null-column revert?

TomyMeren: @L-Lavigne I can answer this.
If you look tot he second part of the where clause, it tries to union inside Charles graph an info from the Bob graph. This generate a null value. 
At revert the changes this query return different result as the expected by Jena

TomyMeren (approved):
LGTM!

L-Lavigne (approved):

PR #686, opened 2022-04-22
Broken rdf testing

TomyMeren:
# Description

## Type of change

- `bug` if it fixes a bug


modules/rdf-tests/src/test/scala/com/gsk/kg/RdfTests.scala
```diff
@@ -60,7 +61,7 @@ class RdfTests extends AnyWordSpec with Matchers with SparkSpec {
     "grouping/manifest.ttl",
     "negation/manifest.ttl",
     "project-expression/manifest.ttl",
-    "property-path/manifest.ttl",
+    //"property-path/manifest.ttl", //time out error
```
L-Lavigne, line 64: Is this timeout a performance issue in Bellman? If so let's create a ticket to address it and put a TODO there.

rdf-test-results.md
```diff
@@ -8,12 +8,13 @@ Overall status
 |-----------|---------------|----------|
 |  Skipped  |       0       |    0     |
 | Disabled  |       0       |    0     |
-|  Failed   |      179      |   100    |
-|**Success**|       0       |    0     |
+|  Failed   |      18       |    69    |
+|**Success**|       8       |    30    |
```
L-Lavigne, line 12: Nice! 👏 

Do you know why the total number of tests has gone down so much? Change in the online test suite contents, or commenting out of `property-path/manifest.ttl` ?

modules/rdf-tests/src/test/scala/com/gsk/kg/RdfTests.scala
TomyMeren: On my local machine it looks like a timeout but I've run it with and action and it's a java.lang.OutOfMemoryError: Java heap space

Here is the full error:

https://github.com/gsk-aiops/bellman/runs/6133214990?check_suite_focus=true

It doesn't look like a performance issue in Belmam, but I put a TODO to check when we go to do the full review of the rdf test

rdf-test-results.md
TomyMeren: Sorry, I had not updated the report. :sweat_smile:

The “Run RDF test" has worked on the accion:
https://github.com/gsk-aiops/bellman/actions/runs/2209578543
I guess the error is from running it manually

modules/rdf-tests/src/test/scala/com/gsk/kg/RdfTests.scala
L-Lavigne: Before ignoring this test, could we try a few more things to fix the error? When googling the logged warning "Broadcasting large task binary" I find answers related to partition sizes like this one, might be worth a try: https://stackoverflow.com/questions/28878654/spark-using-python-how-to-resolve-stage-x-contains-a-task-of-very-large-size-x

TomyMeren: I have found the source of the error but I can't understand the problem. Maybe @patofp can.
The problem is not an error in the input data partition because the input data only has 11 triples.

I first saw that it didn't affect all tests so I took the first one that crashes the test to study it. The problem is not with the query, as other tests use the same query with good results, so the problem is with the input data.

Basically the error occurs when we have a reciprocal relationship of people who know each other, in the input data and the predicate is <http://xmlns.org/foaf/0.1/knows> instead of <http://xmlns.com/foaf/0.1/knows>.

TomyMeren: So for the same query:

SELECT *
WHERE { ?X foaf:knows* ?Y }
ORDER BY ?X ?Y

With that df works:

 val df = List(
            ("<http://example.org/a>", "<http://xmlns.org/foaf/0.1/knows>", "<http://example.org/b>"),
            ("<http://example.org/b>", "<http://xmlns.org/foaf/0.1/knows>", "<http://example.org/c>")
          ).toDF("s", "p", "o")

but with this it is blocked:

val df = List(
            ("<http://example.org/a>", "<http://xmlns.com/foaf/0.1/knows>", "<http://example.org/b>"),
            ("<http://example.org/b>", "<http://xmlns.com/foaf/0.1/knows>", "<http://example.org/c>")
          ).toDF("s", "p", "o")

Obviously changing the foaf definition also in the query

L-Lavigne: That's weird. 🤔 Just curious, any reason why we sometimes use `xmlns.org` instead of `xmlns.com` in our tests and examples? A quick google search shows that only `xmlns.com` should be valid.

TomyMeren: I have changed my conclusion after a review, please read the comments again if you have already done so.

L-Lavigne: Thanks for the update, in your edited comment I see 3 examples now, are you sure the first one works? I would assume that the first 2 would fail, since they both have a "knows" cycle. Can you confirm?

L-Lavigne: I assume our actual data, including the RDF-test data, has the full namespaces and not "foaf:knows", correct? If so, does the bug with cyclic "knows" only happen when using a prefix in the query? In other words, if you use the full namespace in your query, does it work with cyclic data?

patofp: Just to give some insight about property paths. The current implementation of property paths is based on a BFS implementation with Dataframes expanding all the nodes of the graphs to search for the paths that match the query by depth levels (path 0 - > pat 1 -> path 2 -> path n). This implementation has basically two problems.

The first one has to do with performance, expanding a graph to search for paths is very costly in time and space as it has to keep track of explored paths on an auxiliary Dataframe. We can have some discussion about this particular issue this Friday and see the details of the implementation and search for improvements.

The other one has to do with graph cycles, if our data has cycles (KG is a DG but not a DAG) then it can keep creating new paths in a loop, currently the only cycle protection we have is a hard limit that is hard coded. Without deeper analysis and based on the comments, **I think this could be the issue that is affecting**.

This is a cycle in the input dataframe:
```
val df = List(
("http://example.org/a", "foaf:knows", "http://example.org/b"),
("http://example.org/b", "foaf:knows", "http://example.org/a")
).toDF("s", "p", "o")
```

There is a posible alternative implementation for property paths that may be worth to explore [Spark GraphFrames](https://graphframes.github.io/graphframes/docs/_site/index.html), my suggestion is to do a **Property Paths GraphFrame PoC** to see if some of the issues can be solved and if it has other drawbacks as well.

Said that, right now I don't think there is a priority on addressing this known issues, leaving it on behalf of the end user to make a responsable use of property paths.

So IMHO I think that addressing this issue is not a trivial thing that we can do in the scope of this task and should be addressed separately.

Hope this insight helps to take a decision! 😉

TomyMeren: @L-Lavigne The rdf-test uses a ttl file like this:

@prefix : <http://example.org/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

:a foaf:knows :b .
:b foaf:knows :c .

The predicate is define by a "prefix" but the file has the @prefix information to write the data with the full namespace and that is what it does when creating the Dataframe. In conclusion the input Dataframe into the rdf-test always have the full namespace.

For me define the data with prefix in the input Dataframes in any of the unit test is and error because the change to full name occurs earlier and it is not able to match these short predicates with the predicates of the query.

Another point. The bug happens regardless of whether you define the prefix in the query or you write the predicates with the full namespace, does not affect. With this data and this query also fails:

val df = List(
("http://example.org/a", "http://xmlns.com/foaf/0.1/knows", "http://example.org/b"),
("http://example.org/b", "http://xmlns.com/foaf/0.1/knows", "http://example.org/a")
).toDF("s", "p", "o")

SELECT *
WHERE { ?X <http://xmlns.org/foaf/0.1/knows>* ?Y }
ORDER BY ?X ?Y

In any case, I'm with @patofp  that it isn't a priority now.
I think that that we can merge this changes to get a higher percentage into the RDF-test report and create an issue to solve this later.

L-Lavigne: Thanks @patofp and @TomyMeren for the detailed clarification. I agree with making the cyclic bug a separate issue so we can move this current PR forward. 

I do think that lack of support for cycles would qualify as a bug since:

a) relationships such as "knows" have cycles in real-world applications;
b) the RDF tests that are failing are testing for cycles specifically (see the test names);
c) libraries such as Jena support them explicitly (search for "cycles" in https://jena.apache.org/documentation/query/property_paths.html).

@TomyMeren if you agree, could you create a new issue, quote some of our discussions above for context, and replace the "OutOfMemoryError" comment in this PR with a TODO link to the new issue? Thanks!

patofp: > I do think that lack of support for cycles would qualify as a bug since:

Indeed @L-Lavigne. Here is the actual hard limit cycle protection if you guys want to take a look at: https://github.com/gsk-aiops/bellman/blob/c7b906ed371517091a5f8e9da78305f1336220f3/modules/engine/src/main/scala/com/gsk/kg/engine/properties/FuncProperty.scala#L97

TomyMeren: https://github.com/gsk-aiops/bellman/issues/691

L-Lavigne (approved):
Thanks @TomyMeren and @patofp for your analysis of the property-paths issue!

I see the build breaks on the recently-updated microsite task though: https://github.com/gsk-aiops/bellman/runs/6191967760?check_suite_focus=true#step:10:35

I have the same issue in a PR of my own. Let's take a look, I'll be back tomorrow morning.

modules/rdf-tests/src/test/scala/com/gsk/kg/RdfTests.scala
```diff
@@ -60,7 +61,7 @@ class RdfTests extends AnyWordSpec with Matchers with SparkSpec {
     "grouping/manifest.ttl",
     "negation/manifest.ttl",
     "project-expression/manifest.ttl",
-    "property-path/manifest.ttl",
+    //"property-path/manifest.ttl", //Todo: https://github.com/gsk-aiops/bellman/issues/691
```
patofp, line 64: Usually we use TODO in upper case letters.

modules/rdf-tests/src/test/scala/com/gsk/kg/RdfTests.scala
```diff
@@ -154,16 +158,32 @@ class RdfTests extends AnyWordSpec with Matchers with SparkSpec {
...
-      Config.default.copy(formatRdfOutput = true)
-    )
+    val result = ResultSetFactory.result(cleanString)
+    if (result.isResultSet) {
+      val resultSet = result.getResultSet
+      val vars = resultSet.getResultVars.asScala
+      val df = resultSet.asScala.toList
+        .map(solution => vars.map(v => {
+          val sO: Try[String] = Try(solution.get(v).toString)
+          val s = sO.getOrElse("")
```
patofp, line 168: Try using meaningful variable names

patofp (changes requested):

L-Lavigne:
@patofp Have your requested changes been satisfied for this PR and if so, could you approve? Thanks!

patofp (approved):
LGTM!

PR #699, opened 2022-05-03
modified sparql method to be graph inclusive

patofp:
# Description

Changed default config to be graph inclusive for `.sparql()` API

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/syntax/package.scala
```diff
@@ -62,7 +62,7 @@ trait DataFrameSyntax {
       * @return
       */
     def sparql(query: String): DataFrame =
-      sparql(query, Config.default)
+      sparql(query, Config.default.copy(isDefaultGraphExclusive = false))
```
L-Lavigne, line 65: Are we changing the default configuration for all of Bellman, or for this specific method?

If it's for all of Bellman, we need to:
- change the value in `Config.default` itself, not just in this method
- change the default config of the unit tests, and "invert" the logic in all tests that were changing it to `false`
- update the docs in `configuration.md` to specify the new default

If we want to change just this specific method, we need to:
- rename the method to something that hints at the non-default behavior; for example the same source file has a `toTypedDF()` extension that sets `typeDataframe = true` as its name indicates. Perhaps we could call it `sparqlInclusive()` or something.
- ensure we have at least some tests that validate the behavior depending on this property (I see no test changes in this PR, and it passed CI, so there might be a gap in test coverage). 
- update the docs in `configuration.md` and `howto.md` to add a note that this method uses a different default.

In either case, this PR should probably be labelled as a breaking change, as it could change the behavior significantly even if the binary API remains compatible.

L-Lavigne (changes requested):
This seems like a significant change, please see my comment for some things we need to clarify.

modules/engine/src/main/scala/com/gsk/kg/engine/syntax/package.scala
patofp: It is just for the `.sparql()` method. Normally this method is used through the Scala REPL so it should not affect the behaviour of all the engine, just for this interface.

> rename the method to something that hints at the non-default behavior
That's an interesting discussion because maybe we should consider that the default behavior should be "inclusive" instead of "exclusive". In the DataQuery API it is being used in 'inclusive' mode and in Spark shell with the `.sparql()` method we are being asked to modify the default behavior to 'inclusive' as well.

> ensure we have at least some tests that validate the behavior depending on this property
We already have test coverage on both behaviors 'inclusive' and 'exclusive' in `CompilerSpec.scala` although the majority of the test are with exclusive mode as the default behavior. We don't have explicit test on the `sparql()` method that test the behavior. The inclusive/exclusive mode affects on how the default graph is composed when named graphs are defined, but depending on the input Dataframe if the `g` column is set to empty it will consider it belongs to the default graph. It is free to the SPARQL endpoint what mode should implement. I recommend taking a look at: https://blog.metaphacts.com/the-default-graph-demystified it is good post about default graph for a complete explanation.

> update the docs in configuration.md and howto.md to add a note that this method uses a different default.
I agree on this.

modules/site/src/main/docs/compilation.md
```diff
@@ -54,5 +54,5 @@ val result = triple1Result.join(triple2Result, "s")
 ```
 
 This is a simple example of how a **SPARQL** query is compiled into **Spark** code. Under the hood, the compiler performs
-a sequence of [**Phases**](phases.md) like transforming the query into an internal **AST** representation. Finally, this 
+a sequence of [**Phases**](phases) like transforming the query into an internal **AST** representation. Finally, this
```
L-Lavigne, line 57: I had changed those to add the `.md` extension and fix the paths in https://github.com/gsk-aiops/bellman/commit/de23573d1cc785b7f28479a5ee0459b75ca6ad1a. Seems like it was incorrect? IDEA was reporting the original links as wrong.

modules/site/src/main/docs/inference.md
```diff
@@ -9,12 +9,12 @@ permalink: docs/inference
 Inference is a way of **reasoning** so that new explicit knowledge is generated from the implicit knowledge
 contained within the data. The **Bellman** engine is capable of doing this inference in two different ways:
 
-- [**Forward Chaining**](phases/forwardChaining.md)
-- [**Backward Chaining**](phases/backwardChaining.md)
+- [**Forward Chaining**](/bellman/docs/phases/forwardChaining)
```
L-Lavigne, line 12: Is there a reason for having the `/bellman/docs/` prefix for those two links only?

modules/site/src/main/docs/howto.md
```diff
@@ -160,6 +160,9 @@ Bellman also provides an API to trigger the compiler from the **Spark DataFrame*
 kg.sparql(query).show(false)
 ```
 
+**Beware that when using this method, the default behavior of the engine will be inclusive mode.
```
L-Lavigne, line 163: I suggest changing `Beware` to `Note`

modules/engine/src/main/scala/com/gsk/kg/engine/syntax/package.scala
L-Lavigne, line 65: Please update the method's scaladoc as well.

L-Lavigne (approved):
As discussed offline, we can make this alteration for `.sparql()` with proper documentation as a breaking change, but we'll want to further discuss our approach to the default configuration  and what clients of the API are actually expecting.

PR #701, opened 2022-05-06
Loading implicit config from application.conf; removed TypeDataFrame

L-Lavigne:


patofp (commented):
To give some context, using configuration as a default argument was our first approach. It was discussed between Pepe and me, he pointed out that default arguments has some problems, I think that some of the arguments he gave were that they break currying and partial application, they also tend to hide the definition/behavior of the methods thus making it less clear and difficult to debug.

Another thing about make it `implicit`, we know how painful can be sometimes implicits, another possibility is to use a `Kleisli` or a `Reader` monad instead, so the requirement of a configuration is encoded in the type signature of the methods (Though, I wouldn't use it on the external interface methods, such as `Compiler.compile` or `.sparql()` where I think implicit is the way to go, as it would make them harder to use to the final user). 

What are your thoughts @L-Lavigne?

L-Lavigne:
@patofp @TomyMeren I've reverted the contentious `.sparql()` change to preserve the legacy behavior; however I've marked this usage as deprecated in the code and docs. So we should be good with merging once approved, functionality-wise. Could you take another look? Thanks!

patofp (approved):
LGTM! Left a comment about the RDF Tests file

TomyMeren (approved):
LGTM!

PR #712, opened 2022-05-23
Add support contains

patofp:
# Description

Adds support for CONTAINS function.

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ContainsSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+    extends AnyWordSpec
+    with Matchers
+    with SparkSpec
+    with TestConfig {
+
+  import sqlContext.implicits._
+
+  "return expected results from CONTAINS function" when {
+
+    "simple query with FILTER" in {
```
L-Lavigne, line 22:
```suggestion
  "CONTAINS function" should {

    "return expected results using simple query with FILTER" in {
```

L-Lavigne (approved):
👍 , just one test wording suggestion

PR #728, opened 2022-06-17
refactor rdf tests

patofp:
# Description

Refactored rdf-tests

## Type of change

- `bug` if it fixes a bug


modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTestsSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.spark.sql.SparkSession
+import cats.implicits._
+
+class RdfTestsSpec extends AnyWordSpec with Matchers {
+
+  lazy val spark: SparkSession =
+    SparkSession
+      .builder()
+      .master("local")
+      .appName("spark test example")
```
L-Lavigne, line 18: More meaningful name?

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
```diff
@@ -0,0 +1,145 @@
...
+
+object RdfUtils {
+
+  private val sparqlModelPath =
+    "modules/rdf-tests/src/test/resources/rdf-tests/sparql11/data-sparql11/"
+
+  def readQuery(pathString: String): IO[String] = for {
+    path <- IO(Paths.get(pathString.toFilePath))
+    file <- IO(Files.readAllBytes(path))
+    s = new String(file)
```
L-Lavigne, line 34: This can go in the `yield` clause

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
```diff
@@ -0,0 +1,145 @@
...
+          val columns = vars
+            .map(columnName =>
+              df("value")(vars.indexOf(columnName)).alias(columnName)
+            )
+          formatDF(df.select(columns: _*))
+        }
+      } else {
+        IO(sr.getBooleanResult).map { r =>
+          List(
+            "\"" + r + "\"^^<http://www.w3.org/2001/XMLSchema#boolean>"
```
L-Lavigne, line 120: Could be an interpolated string

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTestsSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+    "subquery/manifest.ttl",
+    "syntax-query/manifest.ttl"
+  )
+
+  implicit val sc: SQLContext = spark.sqlContext
+
+  suites.foreach { suite =>
+    RdfTests
+      .evalManifest(
+        suite.replace("/maninfest.ttl", "").capitalize
```
L-Lavigne, line 42: `manifest`, not `maninfest`

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
```diff
@@ -0,0 +1,73 @@
...
+      ?suite mf:result ?result .
+      ?suite mf:action _:action .
+      _:action qt:query ?query .
+      _:action qt:data ?data .
+      ?suite mf:result ?result .
+    }
+    """
+
+  def evalManifest[A](manifest: String)(
+      handler: (String, Either[Throwable, (List[Row], List[Row])]) => IO[A]
```
L-Lavigne, line 37: Is passing a handler with a complex signature really necessary here, since we're only calling this once in the code, with the same handler? We could just integrate the handler in the logic here.

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
```diff
@@ -0,0 +1,73 @@
...
+      }.sequence)
+  }
+
+  def toRowList(df: DataFrame)(implicit sc: SQLContext): IO[List[Row]] = IO {
+    import sc.implicits._
+    (if (df.isEmpty) {
+       List("").toDF("value")
+     } else {
+       df
+     }).collect().toList
```
L-Lavigne, line 71: `()` is unnecessary here

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTestsSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.SparkSession
+import cats.implicits._
+
+class RdfTestsSpec extends AnyWordSpec with Matchers {
+
+  lazy val spark: SparkSession =
+    SparkSession
+      .builder()
```
L-Lavigne, line 16: `()` is unnecessary here

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
```diff
@@ -0,0 +1,145 @@
...
+
+    def toDFResults(sr: SPARQLResult): IO[DataFrame] = {
+      if (sr.isResultSet) {
+        for {
+          resultSet <- IO(sr.getResultSet)
+          vars      <- IO(resultSet.getResultVars.asScala.toList)
+          solutions <- IO(resultSet.asScala.toList)
+          results   <- readSolutions(solutions, vars)
+        } yield {
+          val df = results.toDF()
```
L-Lavigne, line 110: `()` is unnecessary here

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
```diff
@@ -0,0 +1,145 @@
...
+            triple.getSubject.toString,
+            triple.getPredicate.toString,
+            triple.getObject.toString
+          )
+        }
+        .toDF("s", "p", "o")
+    }
+
+    for {
+      input     <- IO(new CollectorStreamTriples())
```
L-Lavigne, line 64: `()` is unnecessary here

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/package.scala
```diff
@@ -0,0 +1,8 @@
+package com.gsk.rdf
+
+package object tests {
+
+  implicit class ToFilePath(s: String) {
```
L-Lavigne, line 5: Is a whole extension class necessary for a simple function we're only calling once in the code?

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
```diff
@@ -0,0 +1,145 @@
...
+      rdfParser <- IO(RDFParser.source(filePath))
+      _         <- IO(rdfParser.parse(input))
+      triples   <- IO(input.getCollected.asScala.toList)
+      df        <- IO(inputToDf(triples))
+    } yield formatDF(df)
+  }
+
+  def readResults(filename: String)(implicit sc: SQLContext): IO[DataFrame] = {
```
L-Lavigne, line 72: That function has a _lot_ of `IO` wrappers in it. Considering we're not doing any error handling, can't we just wrap the whole thing in a single `IO`? Like this:

```scala
  def readResults(filename: String)(implicit sc: SQLContext): IO[DataFrame] = IO {
    import scala.collection.JavaConverters._
    import sc.implicits._

    def readSolution(solution: QuerySolution, vars: Seq[String]): Seq[String] =
      vars.map(v => Option(solution.get(v)).map(_.toString).fold("") { s =>
        val splitSol = s.split('^')
        if (splitSol.length <= 1) s else s""""${splitSol(0)}"^^<${splitSol(2)}>"""
      })

    val sr = ResultSetFactory.result(filename.toFilePath)
    if (sr.isResultSet) {
      val resultSet = sr.getResultSet
      val vars      = resultSet.getResultVars.asScala
      val results   = resultSet.asScala.map(readSolution(_, vars)).toList
      val df        = results.toDF
      val columns   = vars.zipWithIndex.map { case(colName, i) => df("value")(i).alias(colName) }
      formatDF(df.select(columns: _*))
    } else
      List(s""""${sr.getBooleanResult}"^^<http://www.w3.org/2001/XMLSchema#boolean>""").toDF("?_askResult")
  }
```

Other functions in this package could probably benefit from a similar treatment.

L-Lavigne (commented):
Looks good, but please see some comments about style esp. re use of `IO` wrappers. (Also note a typo in `manifest`).

TomyMeren (approved):
LGTM

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
patofp: Tried to escape the backslashes and not possible, it throws an error.

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTestsSpec.scala
patofp: Not sure why is this replace, I removed and seems to work fine

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
patofp: Not easy to integrate with the Scalatest matchers, I don't think is bad to have this type of handler as it will give the ability to use it in different use cases or different test frameworks. What do you think?

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/package.scala
patofp: Do you see this as a problem or bad practice? Do you have any suggestions?

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfUtils.scala
patofp: That is true, right now we are not doing any error handling but I think we should, some of the manifest are failing, e.g: `syntax-query` so I think maybe there is need to do some error handling inside this method, by the moment I've only refactored what was already implemented but the are some problems with this code.

PR #733, opened 2022-07-06
Fix group by

patofp:
# Description

Fixed some issues with group by queries:
- Parsing errors
- Function bindings
- Arithmetic functions on group by functions

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/analyzer/AnalyzerSpec.scala
```diff
@@ -214,4 +235,34 @@ class AnalyzerSpec
...
+
+    result.fold(
+      { error =>
+        error shouldEqual EngineError.AnalyzerError(
+          NonEmptyChain(
+            "found free variables VARIABLE(?O12)"
+          )
+        )
+      },
+      _ => succeed
```
L-Lavigne, line 265: Shouldn't this fail, if we're expecting an error?

L-Lavigne (changes requested):
Looks good, just one possible error in tests

modules/engine/src/test/scala/com/gsk/kg/engine/analyzer/AnalyzerSpec.scala
patofp: Good catch, the test description was wrong, this test should not find unbound variables as the `?c` is used in the GROUP BY statement as a function binding, so in that case it should be allowed to use it.

PR #734, opened 2022-07-07
Fixed config name so it can be properly overridden by applications

L-Lavigne:
See https://github.com/lightbend/config#standard-behavior


patofp (approved):
LGTM!

PR #745, opened 2022-07-27
Construct improved perfomance

TomyMeren:
# Description
Changes the first approach to the evaluation of the CONSTRUCT clause.
This first approach consisted of applying a flatMap to the result of the DataFrame generated after evaluating the WHERE part of the query. This required working with the DataFrame as an RDD making it perform poorly and difficult to paralyse.

The new implementation use un-pivot and Union, two Spark SQL methods.

This are examples of times in some queries of the Perfomance test and others. Everything has been run with a single table approach (/hudi_perf_v8) and hudi: 0.10.1

Best with First approach - Bellman: 2.0.0+166-7ddf6580-SNAPSHOT
  - doc_base => 356325 ms
  - mmk => 438292 ms
  - Adapted GeneCards => 5 min

New approach- Bellman: 2.0.0+170-d40b8ab5-SNAPSHOT
- doc_base => 317582 ms
- mmk => 369996 ms
- Adapted GeneCards=> 2 min

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -652,48 +659,119 @@ object Engine {
...
+                stringValToColumn(rel._2)
+              )
+        }
+
+        val dfKey: DataFrame = dfColsPandOGen
+          .withColumn("s", stringValToColumn(key))
+          .select(col("s"), expr(stackExpr))
+          .distinct
+
+        accDF.union(dfKey)
```
L-Lavigne, line 771:
```suggestion
      (accSpoDF: DataFrame, s: StringVal) =>
        val poList: List[(StringVal, StringVal)] = templateValues(s)

        val stackExpr: String =
          (poList.length +: poList.indices.flatMap(i => List(s"p$i", s"o$i")))
            .mkString("stack(", ", ", ") as (p, o)")

        val poDF: DataFrame = poList.zipWithIndex.foldLeft(dfPRev) {
          (accPoDF: DataFrame, po_i: ((StringVal, StringVal), Int)) =>
            val ((p, o), i) = po_i
            accPoDF
              .withColumn(
                s"p$i",
                stringValToColumn(p)
              )
              .withColumn(
                s"o$i",
                stringValToColumn(o)
              )
        }

        val spoDF: DataFrame = poDF
          .withColumn("s", stringValToColumn(s))
          .select(col("s"), expr(stackExpr))
          .distinct

        accSpoDF.union(spoDF)
```

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -706,8 +784,19 @@ object Engine {
       r: Multiset
   )(implicit
       sc: SQLContext
-  ): M[Multiset] =
-    evaluateConstructPlain(bgp, r).pure[M]
+  ): M[Multiset] = {
+
+    // forward chaining inference use CONSTRUCT internally. In these cases the number of quads in BGP is very large and
+    // the number of rows in the Multiset very short. This is a strange case in compared to the usual use of CONSTRUCT
+    // in normal querys. That is why the evaluation of CONSTRUCT changes in these cases and needs a small adaptation.
```
L-Lavigne, line 791:
```suggestion
    // Forward chaining inference uses CONSTRUCT internally. In this case the number of quads in the BGP is very large
    // and the number of rows in the Multiset is very small. This is an unusual case compared to the use of CONSTRUCT
    // in normal queries. That is why the evaluation of CONSTRUCT changes in these cases and needs a small adaptation.
```

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -706,8 +784,19 @@ object Engine {
...
+    val isForwardChaining: Boolean = forwardChainigOWLAxiomnsAndRDFSEntailments
+      .flatMap(_ match {
+        case Construct(bgp, r) => bgp.quads
+      })
```
L-Lavigne, line 795:
```suggestion
      .flatMap {
        case Construct(bgp, _) => bgp.quads
      }
```

L-Lavigne (changes requested):
Good work, I have some suggested changes but the main one is in the construction of the `stack` expression: removing usage of `indexOf` on a list that's not certain to be an `IndexedSeq`, and clarifying the value names.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -706,8 +782,23 @@ object Engine {
...
+    // in normal queries. That is why the evaluation of CONSTRUCT changes in these cases and needs a small adaptation.
+    val forwardChainingQuads: List[Expr] =
+      forwardChainigOWLAxiomnsAndRDFSEntailments
+        .flatMap { case Construct(bgp, _) =>
+          bgp.quads
+        }
+
+    val isInternalForwardChainingUse: Boolean = bgp.quads
+      .map(forwardChainingQuads.contains(_))
+      .reduce(_ && _)
```
L-Lavigne, line 798: For performance reasons we should use a `Set` instead of a `List` when calling `contains`. (Also, there's some typos in val/def names in this section.)

```suggestion
    val forwardChainingQuads: Set[Expr] =
      forwardChainingOWLAxiomsAndRDFSEntailments
        .flatMap { case Construct(bgp, _) =>
          bgp.quads
        }.toSet

    val usingForwardChaining: Boolean = bgp.quads
      .map(forwardChainingQuads.contains)
      .reduce(_ && _)
```

L-Lavigne (changes requested):
One more suggestion, for the new forward-chaining detection logic.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -707,8 +819,24 @@ object Engine {
...
       sc: SQLContext
-  ): M[Multiset] =
-    evaluateConstructPlain(bgp, r).pure[M]
+  ): M[Multiset] = {
+
+    // Forward chaining inference uses CONSTRUCT internally. In this case the number of quads in the BGP is very large
+    // and the number of rows in the Multiset is very small. This is an unusual case compared to the use of CONSTRUCT
+    // in normal queries. That is why the evaluation of CONSTRUCT changes in these cases and needs a small adaptation.
+    val forwardChainingQuads: Set[Expr] =
+      forwardChainigOWLAxiomnsAndRDFSEntailments.flatMap {
```
L-Lavigne, line 828:
```suggestion
      forwardChainingOWLAxiomsAndRDFSEntailments.flatMap {
```

modules/engine/src/main/scala/com/gsk/kg/engine/inference/ForwardChainingInference.scala
```diff
@@ -84,6 +84,26 @@ object ForwardChainingInference {
       .apply(dagQueries)
   }
 
+  val forwardChainigOWLAxiomnsAndRDFSEntailments = List(
```
L-Lavigne, line 87:
```suggestion
  val forwardChainingOWLAxiomsAndRDFSEntailments: List[DAG[Fix[DAG]]] = List(
```

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/SparkSpec.scala
```diff
@@ -14,7 +14,7 @@ trait SparkSpec extends DataFrameSuiteBase { self: Suite =>
     new SparkConf()
       .setMaster("local[*]")
       .setAppName("test")
-      .set("spark.ui.enabled", "false")
+      .set("spark.ui.enabled", "true")
```
L-Lavigne, line 17: Just curious, why this change? I'm not familiar with that option but it sounds like we wouldn't want any UI for unit testing.

L-Lavigne (commented):
Looks good, just one typo remaining, and I have a question about a test setting.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/SparkSpec.scala
TomyMeren: Yes, that's right. I needed to check the Spark SQL DAG for some unit test locally. 
I don't know the reason why it is not enabled. It could be because it affects performance. I'm going to change it to false just in case.

L-Lavigne (approved):

PR #748, opened 2022-08-04
Join DataFrames of the quads, with the DataFrame of the clause VALUE with commons bindings, before join between them.

TomyMeren:
The current behaviour with the VALUES clause is to join the DataFrame it generates, at the end of the process. This causes us to work with a lot of unnecessary data throughout the whole process and only remove it at the end.

With this approach we try to reduce the number of rows from the beginning, in the DataFrames of the quads with common binding to the VALUES DataFrame. 

## Type of change

- `bug` if it fixes a bug


L-Lavigne (commented):
Looks good, just one syntax clarity suggestion.

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/JoinValues.scala
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  def phase[T](implicit T: Basis[DAG, T]): Phase[T, T] = Phase { t =>
+    val result = apply(T)(t)
+    (result != t)
+      .pure[M]
+      .ifM(
+        Log.debug(
+          "Optimizer(JoinBGPs)",
```
L-Lavigne, line 53:
```suggestion
          "Optimizer(JoinValues)",
```

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardValues.scala
```diff
@@ -0,0 +1,82 @@
...
+            if (allTheCommonValuesAreDef) {
+              // join only with the common and define values
+              val onlyCommonRows = tableRows.map(row =>
+                Row(row.tuples.filter { case (key, _) =>
+                  commonValues.contains(key)
+                })
+              )
+              T.algebra(
+                Join(quad, T.algebra(Table(commonValues, onlyCommonRows)))
+              )
```
L-Lavigne, line 46:
```suggestion
        // try to join the table with the quads whenever they have common variables.
        case Quad(s, p, o, _) =>
          val spo = Set(s, p, o)
          val commonValues = tableVars.filter(spo)
          if (commonValues.isEmpty) {
            quad
          } else if (commonValues.size == tableVars.size) {
            T.algebra(Join(quad, T.algebra(Table(tableVars, tableRows))))
          } else {
            val commonsSet = commonValues.toSet
            // join only with the common and defined values
            val onlyCommonRows = tableRows
              .map(row => Row(row.tuples.filter(t => commonsSet.contains(t._1))))
              .filter(_.tuples.nonEmpty)
            if(onlyCommonRows.size == tableRows.size) {
              T.algebra(Join(quad, T.algebra(Table(commonValues, onlyCommonRows))))
```

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardValues.scala
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  def phase[T](implicit T: Basis[DAG, T]): Phase[T, T] = Phase { t =>
+    val result = apply(T)(t)
+    (result != t)
+      .pure[M]
+      .ifM(
+        Log.debug(
+          "Optimizer(JoinValues)",
```
L-Lavigne, line 74: What is the meaning of "Moving Forward" in this context? It seems unclear.

```suggestion
          "Optimizer(MovingForwardValues)",
```

L-Lavigne (commented):
I made a suggestion to reduce collection traversals a bit.

Also I had a question about the name of this optimization.

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardValues.scala
TomyMeren: This change applies the join between the different quads and the table generated by the values VALUE clause, with common columns, instead of doing the join only once at the end of the process. The idea is to remove unwanted data as soon as possible. 

So it is a way to forward (and also distribute) the joins with the table generated in the VALUE clause in the DAG.

Perhaps a more convenient name could be advanceJoinValueClause

L-Lavigne: So by "forward", do you mean "forward in time", as in "earlier"? If so, I think a name like "EarlyJoinOnValueClause" would be more clear. Please make sure to update the name in the debug statement also.

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -828,6 +828,7 @@ object Engine {
       forwardChainingOWLAxiomsAndRDFSEntailments.flatMap {
         case Construct(bgp, _) =>
           bgp.quads
+        case _ => Nil
```
vil1, line 831: Is this for removing the warning about pattern-matching exhaustiveness checking?

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -881,7 +882,7 @@ object Engine {
 
     Multiset(
       bindings = (vars :+ VARIABLE(GRAPH_VARIABLE.s)).toSet,
-      dataframe = df
+      dataframe = broadcast(df)
```
vil1, line 885: I can't work out why this `broadcast` is useful here (bu I'm not very good at spark)

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValue.scala
```diff
@@ -0,0 +1,95 @@
...
+
+import com.gsk.kg.config.Config
+import com.gsk.kg.config.InferenceMode
+import com.gsk.kg.engine.DAG._
+import com.gsk.kg.engine.data.ToTree._
+import com.gsk.kg.sparqlparser.Expr.Row
+import com.gsk.kg.sparqlparser.Result
+import com.gsk.kg.sparqlparser.StringVal
+
+object EarlyJoinOnValue {
```
vil1, line 19: Since we already have `PushDownBGPs`, maybe we could name this `PushDownJoins` or `PushDownJoinOnValue`?

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
TomyMeren: Yes, it is true that it is not related to this PR.  I can separate it into another PR

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
TomyMeren: Broadcast sends a copy of the DataFrame to all nodes in the cluster to join the data separately. This greatly improves performance when we join two DataFrames, one large and one small, and apply bradcast to the small one. Especially in parallelism.

In this case the table generated in the VALUES clause is created manually in the query so we can say that it will always be smaller than the input DataFrame.

Translated with www.DeepL.com/Translator (free version)

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValue.scala
TomyMeren: PushDownJoinOnValue sounds good, @L-Lavigne what do you think?

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValue.scala
```diff
@@ -0,0 +1,95 @@
...
+        case _ => quad
+      }
+    )
+  )
+
+  def apply[T](config: Config)(implicit T: Basis[DAG, T]): T => T = { t =>
+    T.coalgebra(t).rewrite { case j @ Join(l, r) =>
+      // TODO: backward chaining inference is not ready for BGPs with filters inside it and needs adaptation
+      //  https://github.com/gsk-aiops/bellman/issues/761
+      if (
```
vil1, line 61: I think it would be better to put this check directly in the `phase` (and therefore we wouldn't need the `config` parameter here anymore)

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ValuesSpec.scala
```diff
@@ -312,42 +321,5 @@ class ValuesSpec
         )
       )
     }
-
```
vil1, line 315: It isn't clear to me why this test isn't needed anymore

vil1 (changes requested):
Looks good overall, but with minor comments nonetheless

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValue.scala
L-Lavigne: I'm not very familiar with the Spark "push down" concept and whether it's the correct name for our optimizations, but if it is, then I'm OK with using the name here and also in https://github.com/gsk-aiops/bellman/pull/759.

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValueSpec.scala
```diff
@@ -0,0 +1,594 @@
...
+                    )
+                  )
+                )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+
+      "using a two variable in the clause, one appearing in a single quad and the other in more than one quad" in {
```
L-Lavigne, line 215:
```suggestion
      "using two variables in the clause, one appearing in a single quad and the other in more than one quad" in {
```
(same applies to other tests with similar descriptions)

L-Lavigne (commented):
Looks good, just some typos in tests.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ValuesSpec.scala
TomyMeren: I have added a new test and removed two that were repeated.

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValue.scala
TomyMeren: It is not a spark concept, if anything by Jena Fuseki, but in any case the important thing is that it is understood.
The other PR is already merged so I'll keep the name and change it later, with other little things in a new PR.

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyJoinOnValue.scala
TomyMeren: Done!

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
vil1: I think it can remain there, my question was just out of curiosity

TomyMeren:
I have made a general re-naming of part of the optimised ones and removed the apply to phase setting in the two new optimised ones.

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/PushDownJoinOnValue.scala
```diff
@@ -0,0 +1,95 @@
...
+        case (Table(tableVars, tableRows), BGP(quads)) =>
+          addJoinValues(quads, tableVars, tableRows)
+        case (BGP(quads), Table(tableVars, tableRows)) =>
+          addJoinValues(quads, tableVars, tableRows)
+        case _ => j
+      }
+    }
+  }
+
+  def phase[T](implicit T: Basis[DAG, T]): Phase[T, T] = {
```
L-Lavigne, line 69: This function looks like a copy of the one in PushDownFilter. Can we refactor those two into one?

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/PushDownJoinOnValueSpec.scala
```diff
@@ -0,0 +1,594 @@
...
+                    )
+                  )
+                )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+
+      "using a two variable in the clause, one with undefined values, one appearing in a single quad and the other in more than one quad" in {
```
L-Lavigne, line 334:
```suggestion
      "using two variables in the clause, one with undefined values, one appearing in a single quad and the other in more than one quad" in {
```
Again, this applies to all similar test descriptions.

L-Lavigne (changes requested):

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/PushDownJoinOnValueSpec.scala
TomyMeren: Fixed in two other places

L-Lavigne (changes requested):

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/PushDownJoinOnValue.scala
TomyMeren: I’ve done a refactor. Let me know if it ’s correct

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/Optimizer.scala
```diff
@@ -1,11 +1,67 @@
...
+          (result != t)
+            .pure[M]
+            .ifM(
+              Log.debug(
+                s"Optimizer(${this.getClass})",
+                s"resulting query: ${result.toTree.drawTree}"
+              ),
+              ().pure[M]
+            ) *>
+            result.pure[M]
```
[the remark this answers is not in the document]
TomyMeren: Done!

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/PushDownJoinOnValueSpec.scala
```diff
@@ -0,0 +1,594 @@
...
+                    )
+                  )
+                )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+
+      "using a two variable in the clause, the both with undefined values, one appearing in a single quad and the other in more than one quad" in {
```
L-Lavigne, line 403:
```suggestion
      "using two variables in the clause, both with undefined values, one appearing in a single quad and the other in more than one quad" in {
```

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/PushDownJoinOnValueSpec.scala
```diff
@@ -0,0 +1,594 @@
...
+                    )
+                  )
+                )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+
+      "using a two variable in the clause, the both with defined and undefined values, one appearing in a single quad and the other in more than one quad" in {
```
L-Lavigne, line 483:
```suggestion
      "using two variables in the clause, both with defined and undefined values, one appearing in a single quad and the other in more than one quad" in {
```

L-Lavigne (commented):
Almost there, just some typos left in test descriptions

L-Lavigne (approved):

vil1 (approved):

PR #758, opened 2022-09-05
Implement casts to XSD datatypes

vil1:
# Description

This PR adds parsing support and compilation for casting values to typed literal, by the means of XSD constructors, such as `xsd:string` in the following query:

```
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
SELECT
	?a
	?v
	(xsd:string(?v) AS ?string)
WHERE {
    ?a :p ?v .
}
```

## Type of change

Functionality implementation.

- `bug` if it fixes a bug


L-Lavigne (approved):
Looks good, congrats @vil1 on your first Bellman PR! 👍

PR #759, opened 2022-09-06
Filter improved perfomance

TomyMeren:
# Description

New Optimisation step. This optimisation applies the filters to the individual quads in order to move the filters forward in the DAG and remove unused data as early as possible. 
For this all filter variables have to be declared in the quad. In case the filter cannot be applied to any of the individual quads, it will be applied to the BGP with all the variables at the end of the process, as the non-optimised tree.
Ex:
Original DAG
Filter(BGP(quad1_title, quad2_price, quad3_author), graterThan(price, 20))

Optimised DAG
BGP(quad1_title, Filter(quad2_price, graterThan(price, 20)), quad3_author)

A next step, pending to implement, could be to apply the filters to a group of quads (BGPs), without breaking the order of the quads, when we have related enough of them to apply the filter instead of waiting to have them all related at the end.
Ex:
Original DAG
Filter(BGP(quad1_title, quad2_author, quad3_price, quad4_priceWithDiscount), equal(price, priceWithDiscount ))

Optimised DAG 
BGP(quad1_title, quad2_author, Filter(BGP(quad3_price, quad4_priceWithDiscount), equal(price, priceWithDiscount )))

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardFilter.scala
```diff
@@ -0,0 +1,110 @@
...
+import com.gsk.kg.engine.data.ToTree._
+import com.gsk.kg.sparqlparser.Conditional.AND
+import com.gsk.kg.sparqlparser._
+import higherkindness.droste.Basis
+
+/**
+  * This optimisation applies the filters to the individual quads in order to move the filters forward in the DAG and
+  * remove unused data as early as possible. For this all filter variables have to be declared in the quadrant.
+  * In case the filter cannot be applied to any of the individual quadrants, it will be applied to the BGP at the end
+  * of the process, as the non-optimised tree.
```
L-Lavigne, line 17:
```suggestion
 * This optimization applies `Filter`s to the individual quads in order to move the filters forward in the DAG and
  * remove unused data as early as possible. For this all filter variables have to be declared in the quad.
  * In case the filter cannot be applied to any of the individual quads, it will be applied to the BGP at the end
  * of the process, as the non-optimized path.
```

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardFilter.scala
```diff
@@ -0,0 +1,110 @@
...
+        calculateFilterStatus[T](q, filter) match {
+          case FilterStatus.Applicable => status.copy(applicable = status.applicable :+ filter)
+          case FilterStatus.NotApplicable => status.copy(notApplicable = status.notApplicable :+ filter)
+        }
+    }
+    val filteredQuad = NonEmptyList
+      .fromList(filterStatusLists.applicable)
+      .fold(q)(filters => T.algebra(Filter(filters, q)))
+
+    FilteredQuadResult(filteredQuad, filterStatusLists.notApplicable)
```
L-Lavigne, line 81:
```suggestion
  def calculateFilteredQuads[T](q: T, filters: NonEmptyList[Expression])(implicit T: Basis[DAG, T]): FilteredQuadResult[T] = {
    val filtersWithChildrenOfAND = filters.flatMap(withChildrenOfAND)
    val (applicableFilters, nonApplicableFilters) =
      filtersWithChildrenOfAND.toList.partition(calculateFilterStatus(q, _) == FilterStatus.Applicable)
    val filteredQuad = NonEmptyList
      .fromList(applicableFilters)
      .fold(q)(filters => T.algebra(Filter(filters, q)))
    FilteredQuadResult(filteredQuad, nonApplicableFilters)
  }
```

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardFilter.scala
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  //the AND logical operator is the only Expression that can optimized even if it contains variables of different quad
+  //because can split the two children expression in two elements of the filterList without changing behaviour
+  def andToList(filterList: NonEmptyList[Expression]):NonEmptyList[Expression] = {
+    filterList.flatMap(filter => filter match {
+      case AND(l, r) => andToList(NonEmptyList.one(l)) ::: andToList(NonEmptyList.one(r))
```
L-Lavigne, line 88:
```suggestion
  //The AND logical operator is the only Expression that can optimized even if it contains variables of different quads,
  //because we can split the two child expressions in two elements of the filter list without changing behaviour
  def withChildrenOfAND(filter: Expression): NonEmptyList[Expression] = {
    filter match {
      case AND(l, r) => withChildrenOfAND(l) ::: withChildrenOfAND(r)
```

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardFilter.scala
```diff
@@ -0,0 +1,110 @@
...
+      case _ => NonEmptyList.one(filter)
+    }
+    )
+  }
+
+  def calculateFilterStatus[T](q: T, f: Expression): FilterStatus = q match {
+      case quad: Quad[_] =>
+        val filterVariables: List[StringVal.VARIABLE] = calculateFilterVariables(f)
+        val variablesInQuad = filterVariables.filter(List(quad.s, quad.p, quad.o).contains(_))
+        if (filterVariables.size == variablesInQuad.size) {
```
L-Lavigne, line 98:
```suggestion
      case Quad(s, p, o, _) =>
        val spo = Set(s, p, o)
        if (findVariablesIn(f).forall(spo.contains)) {
```

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/MovingForwardFilter.scala
```diff
@@ -0,0 +1,110 @@
...
+          FilterStatus.Applicable
+        } else {
+          FilterStatus.NotApplicable
+        }
+      case _ => FilterStatus.NotApplicable
+    }
+
+  //Deepening the children of the expression by looking for variables
+  def calculateFilterVariables(f: Expression): List[StringVal.VARIABLE] = f.getInnerStringValList.collect {
```
L-Lavigne, line 107:
```suggestion
  //Recursively traverse the children of the expression looking for Variables
  def findVariablesIn(f: Expression): List[StringVal.VARIABLE] = f.getInnerStringValList.collect {
```

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/SparkSpec.scala
```diff
@@ -14,7 +14,7 @@ trait SparkSpec extends DataFrameSuiteBase { self: Suite =>
     new SparkConf()
       .setMaster("local[*]")
       .setAppName("test")
-      .set("spark.ui.enabled", "false")
+      .set("spark.ui.enabled", "true")
```
L-Lavigne, line 17: Looks like this got changed again by mistake

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/FilterSpec.scala
```diff
@@ -264,6 +297,59 @@ class FilterSpec
...
+            |   ?book dc:title ?title ;
+            |         ns:price ?price .
+            |   FILTER(?price > 26 || ?title > 22) FILTER(?book = 4)
+            |}
+            |""".stripMargin
+
+        val result = Compiler.compile(df, None, query, config)
+
+        result shouldBe a[Right[_, _]]
+        result.right.get.collect//.length shouldEqual 2
```
L-Lavigne, line 343: Was this left commented out by accident? We should be looking at the contents of the result.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/FilterSpec.scala
```diff
@@ -48,6 +48,8 @@ class FilterSpec
         result shouldBe a[Right[_, _]]
         result.right.get.collect.length shouldEqual 1
         result.right.get.collect.toSet shouldEqual Set(Row("\"Blank\""))
+
+        //Thread.sleep(300000)
```
L-Lavigne, line 52: This can be removed

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
```diff
@@ -5,181 +5,448 @@ import com.gsk.kg.sparqlparser.StringVal.URIVAL
 /** @see
   *   Syntax: [[https://www.w3.org/TR/sparql11-query/#rExpression]]
   */
-sealed trait Expression
+sealed trait Expression {
+  def getInnerStringValList:List[StringVal]
```
L-Lavigne, line 9:
```suggestion
  def getInnerStringValList: List[StringVal]
```

L-Lavigne: Is this method expected to be called more than once for a given `Expression` instance? If so, I suggest making the implementations `lazy val`s instead of `def`s.

L-Lavigne (changes requested):
Good work, please see my suggestions and questions.

Also the same applies as my comment in the other "moving forward" PR (https://github.com/gsk-aiops/bellman/pull/748#discussion_r952231115), I think we could use a clearer name.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/FilterSpec.scala
TomyMeren: Yes... my bad

modules/parser/src/main/scala/com/gsk/kg/sparqlparser/Expression.scala
L-Lavigne: Yes, exactly 👍

TomyMeren: Changed

modules/engine/src/main/scala/com/gsk/kg/engine/optimizer/EarlyFilter.scala
```diff
@@ -0,0 +1,142 @@
...
+      case AND(l, r) => withChildrenOfAND(l) ::: withChildrenOfAND(r)
+      case _         => NonEmptyList.one(filter)
+    }
+  }
+
+  def calculateFilterStatus[T](q: T, f: Expression): FilterStatus = q match {
+    case Quad(s, p, o, _) =>
+      val spo = Set(s, p, o)
+      if (
+        !findVariablesIn(f).isEmpty && findVariablesIn(f).forall(spo.contains)
```
L-Lavigne, line 128:
```suggestion
      val variables = findVariablesIn(f)
      if (
        variables.nonEmpty && variables.forall(spo.contains)
```

L-Lavigne (changes requested):
Looks good, just one more change to make, to memoize a computation result.

L-Lavigne (approved):

PR #764, opened 2022-09-14
Fix the way test data is fetched by also fetching graphs defined with…

vil1:
… mf:graphData

# Description

Fix the way test suites and data are retrieved, by also loading the non-default graphs specified with `mf:graphData` and also making the default graph (specified with `mf:data`) optional, which makes available some tests that were ignored with the previous query.

This PR also changes the sbt config so that a new JVM is forked, with specific arguments, for running the tests under `rdf-tests`.

## Type of change

- `bug` if it fixes a bug


modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
```diff
@@ -47,13 +51,25 @@ object RdfTests {
       .flatMap(_.map { solution =>
         IO(solution.get("?name").toString).flatMap { nameVar =>
           val exec = for {
-            queryVar        <- IO(solution.get("?query").toString)
-            dataVar         <- IO(solution.get("?data").toString)
-            resultVar       <- IO(solution.get("?result").toString)
+            queryVar <- IO(solution.get("?query").toString)
```
L-Lavigne, line 54: Not directly related to this PR, but as I mentioned in a past review for this code I'm not a fan of having every statement wrapped in IO when we're not doing local error handling. Possibly I'm missing something, but it seems to me that the whole block could be inside a single IO. Also, there's lots of repetition in the form ` IO(solution.get("?[...]").toString)` which could be refactored into a single function.

Anyway, not a blocker for the current PR.

L-Lavigne (approved):
Looks good 👍 , though please see my comment about the `IO` functions; might be worth discussing as a team for later.

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
vil1: I think you're right, so I'll introduce such function (and run scalafmt as well :) )

TomyMeren (approved):
LGTM!

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
```diff
@@ -38,40 +43,102 @@ object RdfTests {
...
-      }.sequence)
+      }
+  }
+
+  def computeExpectedAndActualResults(
+      manifest: String,
+      suite: SuiteQuerySolution
+  )(implicit
+      sc: SQLContext
+  ): IO[Either[Throwable, (List[Row], List[Row])]] = {
```
L-Lavigne, line 72: `IO` has an instance of `MonadError` so we shouldn't need to wrap an `Either[Throwable, ...]`, we can use `raiseError` and similar methods for error handling.

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTests.scala
```diff
@@ -38,40 +43,96 @@ object RdfTests {
...
+    ], String, String, String, SuiteQuerySolution](
+      Either
+        .fromOption(extractVariable("?name"), MissingInSolution("?name"))
+        .toEitherNel,
+      Either
+        .fromOption(extractVariable("?query"), MissingInSolution("?query"))
+        .toEitherNel,
+      Either
+        .fromOption(extractVariable("?result"), MissingInSolution("?result"))
+        .toEitherNel
```
L-Lavigne, line 115: These 3 similar elements could also be good candidates for refactoring into a utility function

vil1: Yeah, I considered it and then thought "meh, that's just a test". Also,  I don't forsee any new variable being introduced here.

L-Lavigne: Fair enough :)

L-Lavigne (approved):

PR #765, opened 2022-09-15
Xsd constructors

vil1:
# Description

This PR adds parsing support and compilation for casting values to typed literal, by the means of XSD constructors, such as `xsd:string` in the following query:

```
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
SELECT
	?a
	?v
	(xsd:string(?v) AS ?string)
WHERE {
    ?a :p ?v .
}
```

When a cast fails, a `null` value is output in the corresponding column (instead of the `""` expected in the `rdf-tests`).

## Type of change

Functionality implementation.

- `bug` if it fixes a bug


TomyMeren (approved):
LGTM!

modules/rdf-tests/src/test/scala/com/gsk/rdf/tests/RdfTestsSpec.scala
```diff
@@ -51,14 +51,70 @@ class RdfTestsSpec extends AnyWordSpec with Matchers {
...
+  def compareRows(expected: Row, actual: Row): Option[RowDiff] = {
+    if (expected.schema.fields.length != actual.schema.fields.length) {
+      Some(RowDiff(expected, actual))
+    } else {
+      expected.schema.fieldNames.foldLeft(Option.empty[RowDiff]) {
+        (diff, field) =>
+          val idx = expected.schema.fieldIndex(field)
+          if (!compareFields(expected.get(idx), actual.get(idx))) {
+            diff.orElse(Some(RowDiff(expected, actual)))
+          } else { diff }
```
L-Lavigne, line 76: The handling of diffs seems a bit complex. This could be simplified to:
```suggestion
      val diff = expected.zip(actual).flatMap { case(expRow, actRow) =>
        if(!compareRows(expRow, actRow)) {
          Some(RowDiff(expRow, actRow))
        } else { None }
      }
      diff shouldBe Nil
    }
  }

  def compareRows(expected: Row, actual: Row): Boolean = {
    if (expected.schema.length != actual.schema.length) {
      false
    } else {
      expected.schema.fieldNames.forall { field =>
        compareFields(field, actual.get(expected.schema.fieldIndex(field)))
```

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/EarlyFilterSpec.scala
```diff
@@ -0,0 +1,889 @@
...
+                      )
+                    )
+                  )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+
+      "are diferent and coincident one in one quad and the other in the both" in {
```
L-Lavigne, line 620: Sorry I can't make any sense of this description, could you rewrite it?

Also typo in `different`

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/EarlyFilterSpec.scala
```diff
@@ -0,0 +1,889 @@
...
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+    }
+
+    "two filter clauses variables" when {
+
+      "are different and coincident each one in one quad" in {
```
L-Lavigne, line 556: Not sure what the last part of this description means, could you clarify?

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/EarlyFilterSpec.scala
```diff
@@ -0,0 +1,889 @@
...
+                      )
+                    )
+                  )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+
+      "one split and each one of his part coincidet in one quad. The other is not coincident" in {
```
L-Lavigne, line 756: I also don't understand this description. "_his_ part"? Also typo in the first `coincident`

modules/engine/src/test/scala/com/gsk/kg/engine/optimizer/EarlyFilterSpec.scala
```diff
@@ -0,0 +1,889 @@
...
+                  )
+              case _ => fail
+            }
+          }
+          .getOrElse(fail)
+      }
+    }
+
+    "filter without variables" when {
+      "filter without variables" in {
```
L-Lavigne, line 833: Duplicate description?

L-Lavigne (changes requested):
Please see my comments on some test descriptions. Please also see my suggestions from my previous review (re: handling of diffs, and possible use of extractors) in case they were missed. Thanks!

L-Lavigne:
I just updated my comments on test descriptions to be more constructive, please take a look. Thanks!

L-Lavigne (approved):

PR #774, opened 2022-09-23
Various fixes around numerical and temporal functions

vil1:
# Description

Raise the success rate of rdf-tests to 60% by:
- making math functions (floor, ceil, round, abs) return the same type as their input
- fixing the name of `xsd:dateTime` (previously misspelled as `xsd:datetime`
- truncating the leading zeros for the output of temporal accessors (hours, minutes, etc)

## Type of change

- `bug` if it fixes a bug


modules/engine/src/main/scala/com/gsk/kg/engine/typed/functions/FuncDates.scala
```diff
@@ -19,11 +19,20 @@ object FuncDates {
       date_format(current_timestamp, "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
     )
 
-  def year(col: Column): Column = apply(sYear, col)
+  def year(col: Column): Column = {
+    val Year = 0
```
L-Lavigne, line 23: I see this was started before this PR, but it's weird to declare constants locally like this, especially since one of them (`Seconds`) is declared in the object. Can we move the others to the object?

modules/engine/src/main/scala/com/gsk/kg/engine/typed/functions/TypedLiterals.scala
```diff
@@ -203,4 +203,18 @@ object TypedLiterals {
         )
       )
   }
+
+  implicit class StringToTypedOps(str: String) {
+
+    def asTyped(tpe: String): String =
+      s""""${str}"^^<http://www.w3.org/2001/XMLSchema#${tpe}>"""
```
L-Lavigne, line 210: Curly brackets are not needed here

L-Lavigne (approved):

PR #780, opened 2022-09-28
Implement parsing/writing of RDF* triples + IsTriple function. sbt 1.7.1

L-Lavigne:
Part of https://myjira.gsk.com/browse/KNGRPH-1376

This only implements reading/writing of `<<quoted triples>>` and one of the SPARQL-* functions (IsTriple), to support kg-dataQuery operation. Much remains to implement the full spec.

Also upgrades to sbt 1.7.1 by fixing the missing definition of scalaVersion for Scalafix.


vil1 (approved):
This looks pretty good and seems perfectly fine as a first iteration.

However, I have two questions about the next steps:
- after the typing, the content of quoted triples is put unparsed, as a string, in the column's `value` field. I believe it will prove inefficient down the road, since operations on such values would require further parsing of the value. Do you have a plan in mind to remedy this? 
- what about nested quoted triples ? This question obviously factors with the previous one

TomyMeren (approved):
LGTM!
I will remove the simple PR from the sbt update.

L-Lavigne:
> * after the typing, the content of quoted triples is put unparsed, as a string, in the column's `value` field. I believe it will prove inefficient down the road, since operations on such values would require further parsing of the value. Do you have a plan in mind to remedy this?
> * what about nested quoted triples ? This question obviously factors with the previous one

That's right, the intent is to have a more robust, recursive solution later when we start implementing the various SPARQL functions on quoted triples. For now, my focus is more on having RDF-* content pass the basic tests from the internal kg-dataQuery-spark repo.

PR #744, opened 2022-07-19
fixed VALUES when UNDEF values

patofp:
# Description

Fix bindings manifest:
- Fixed bug on join, now it takes into account null values when joining

## Type of change

Make sure you add the correct label for the PR:

- `breaking-change` if this PR introduces a breaking change
- `enhancement` if it adds a new feature
- `bug` if it fixes a bug
- `test` if it adds or updates tests
- `documentation` if it adds or updates documentation
- `dependency-update` if it updates dependencies

Issue #700, opened 2022-05-05
Adapt the behaviour of the null value joins 


TomyMeren:
https://myjira.gsk.com/browse/AIPL-5089
fix ignore tests in GraphSpec and JoinSpec

PR #790, opened 2022-10-10
modify the join expression and unignore test

TomyMeren:
# Description

In the the PR https://github.com/gsk-aiops/bellman/pull/744 we fixed some aspect of the behaviour of join null values in SPARQL but the join condition was too much open and create duplicates with queries with related values such as ?s1 in this case

```
SELECT ?s ?p ?o ?s1
WHERE
{
   { ?s1 ?p  ?o .
    ?s ?p ?s1 . }

```
Also we fix some ignore test related with the issue https://github.com/gsk-aiops/bellman/issues/700.

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/GraphSpec.scala
```diff
@@ -1598,6 +1596,12 @@ class GraphSpec
             "\"Charles\"",
             "<http://example.org/charles>"
           ),
+          Row(
+            "_:a",
+            "<mailto:bob@oldcorp.example.org>",
+            null,
+            "<http://example.org/charles>"
+          ),
```
TomyMeren, line 1604: This test ran successfully, but I tried again in Jena-Fuseki and the expected result was wrong.

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/GraphSpec.scala
```diff
@@ -1841,6 +1842,11 @@ class GraphSpec
             "_:a",
             "<mailto:bob@oldcorp.example.org>",
             "\"Charles\""
+          ),
+          Row(
+            "_:a",
+            "<mailto:bob@oldcorp.example.org>",
+            null
```
TomyMeren, line 1849: The same case as before but in this case the result is succesful

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/ReasoningSpec.scala
```diff
@@ -2185,7 +2185,14 @@ class ReasoningSpec
             config.copy(inferenceMode = BackwardChaining)
           )
 
+          /* TODO: The intermediate level of subPropertyOf,<http://example.org/prop/hasHeight>,
+           * creates a duplicate in the result via a null value.
+           */
```
TomyMeren, line 2190: This bug benefited from the previous wrong implementation of join but reveals a problem with BackwardChaining inference.

L-Lavigne (approved):
Thanks a lot for this fix and your other findings @TomyMeren ! I wasn't able to test a local-publish myself because of the other issues I mentionned today, but we're good for a merge and we'll do another PR if I uncover issues later.

vil1 (approved):
LGTM 👍

PR #806, opened 2022-10-26
New JSON-based struct for RDF* triples; implement s/p/o SPARQL* getters

L-Lavigne:
https://myjira.gsk.com/browse/KNGRPH-1504


modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
```diff
@@ -930,4 +883,51 @@ object Engine {
...
+      case BOOL(b)        => defaultStructForLit(lit(b), Boolean)
+      case GRAPH_VARIABLE => defaultStructForLit(lit("*g"), String)
+      case URIVAL(s) =>
+        defaultStructForLit(
+          substring_index(substring_index(lit(s), "<", -1), ">", 1),
+          Uri
+        )
+      case TRIPLE(subj, pred, obj) =>
+        val triple = struct(
+          stringValToColumn(subj).as(sCol),
```
vil1, line 926: Label me a fanatic, but direct recursion feels wrong in this project. There should be a way to rewrite this function using recursion schemes.

modules/engine/src/test/scala/com/gsk/kg/engine/typed/functions/FuncTermsSpec.scala
```diff
@@ -612,5 +593,50 @@ class FuncTermsSpec
...
+      "have SUBJECT() return the outer subject of a nested triple " +
+        "and then have SUBJECT() return its inner subject" in {
+          test(
+            FuncTerms.subject _ andThen FuncTerms.subject,
+            nestedTriple -> s,
+            triple       -> "null",
+            nonTriple    -> "null"
+          )
+        }
+    }
```
vil1, line 640: As far as I understand, the object of a RDF-* triple can be a (nested) quoted triple. If so, can you add a test case for that.

Similarly (again, AFAIU), the predicate of a RDF-* cannot be a quoted triple, is there a test showing that doing so is actually considered as an error by the engine?

scalastyle-config.xml
```diff
@@ -103,7 +103,7 @@
  </check>
  <check level="warning" class="org.scalastyle.scalariform.NumberOfMethodsInTypeChecker" enabled="true">
   <parameters>
-   <parameter name="maxMethods"><![CDATA[80]]></parameter>
+   <parameter name="maxMethods"><![CDATA[100]]></parameter>
```
vil1, line 106: Maybe that was more a hint that we should start splitting `Engine` into smaller components :) 
Joking aside, that would help reducing compilation times when we do some changes on it.

vil1 (changes requested):
This is definitely going in the right direction, although I have concerns about the new implementation of `stringValToColumn`

modules/engine/src/main/scala/com/gsk/kg/engine/Engine.scala
L-Lavigne: Makes sense, though I've never used recursion schemes before so I'll need some help figuring this out. I'll be reaching out to set up a discussion.

modules/engine/src/test/scala/com/gsk/kg/engine/typed/functions/FuncTermsSpec.scala
L-Lavigne: Sounds good, I'll add those tests.

scalastyle-config.xml
L-Lavigne: Yeah this is a temporary workaround, the engine could certainly use a good cleanup.

L-Lavigne:
@vil1 @TomyMeren Any idea why the docker process for `build-microsite` keeps returning exit code 1 and failing CI? I don't see any error message from its output.

TomyMeren:
No idea but @vil1 has the same problem on his PR  #807. It's strange.
Have you try to build the microsite locally?

modules/engine/src/main/scala/com/gsk/kg/engine/data/ToTree.scala
```diff
@@ -201,6 +201,9 @@ object ToTree extends LowPriorityToTreeInstances0 {
           case ExpressionF.ISURI(s)     => Node("ISURI", Stream(s))
           case ExpressionF.ISIRI(s)     => Node("ISIRI", Stream(s))
           case ExpressionF.ISTRIPLE(s)  => Node("ISTRIPLE", Stream(s))
+          case ExpressionF.SUBJECT(s)   => Node("ISTRIPLE", Stream(s))
+          case ExpressionF.PREDICATE(s) => Node("PREDICATE", Stream(s))
+          case ExpressionF.OBJECT(s)    => Node("ISTRIPLE", Stream(s))
```
TomyMeren, line 206: Is the node relationship of SUBJECT and OBJECT correct?

L-Lavigne: Oops, good catch. Will fix

modules/engine/src/main/scala/com/gsk/kg/engine/RdfFormatter.scala
```diff
@@ -107,8 +107,8 @@ object RdfFormatter {
       column.startsWith("mailto:") ||
       column.startsWith("file://")
 
-  def isTriple(column: Column): Column =
```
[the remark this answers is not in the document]
L-Lavigne: I just renamed it `isJson` when switching to that format. Come to think of it, I should add some validation that the JSON structure contains s/p/o, and rename it `isTriple`.

L-Lavigne:
> No idea but @vil1 has the same problem on his PR #807. It's strange. Have you try to build the microsite locally?

I tried and it failed because I didn't have the `dot` program (I think). I'll set that up and try again, since my attempts at solving it through PRs haven't worked so far.

vil1:
I've made a test PR (with a single character change on the README) https://github.com/gsk-aiops/bellman/pull/808 and it fails on build-microsite as well. This probably means that `rlespinasse/drawio-export:latest` introduced the defect (it has been released a week ago). We should find a way to pin the build to the previous version (`4.5.0`).

L-Lavigne:
> I've made a test PR (with a single character change on the README) #808 and it fails on build-microsite as well. This probably means that `rlespinasse/drawio-export:latest` introduced the defect (it has been released a week ago). We should find a way to pin the build to the previous version (`4.5.0`).

Ah, that makes sense, thanks! Let's look into locking that dependency.

vil1:
A disclaimer though, I tried setting the image name to `rlespinasse/drawio-export:4.5.0` in `Diagrams.scala` and it didn't work.

TomyMeren:
It seems that the error start with the merge of the PR #800

PR #807, opened 2022-10-27
Type promotion

vil1:
# Description

Raise success rate of rdf-test to 72% by improving implementation of most of the aggregation functions, namely:
- `MIN`, `MAX`: perform comparison on numbers (when relevant) and not lexicographically, return properly typed results
- `SUM`, `AVG`: implement type promotion on the result

## Type of change

- `bug` if it fixes a bug


modules/engine/src/test/scala/com/gsk/kg/engine/compiler/GroupBySpec.scala
```diff
@@ -582,15 +582,15 @@ class GroupBySpec
         result.right.get.collect.toSet shouldEqual Set(
           Row(
             "<http://uri.com/subject/a1>",
-            "\"3.0\"^^<http://www.w3.org/2001/XMLSchema#double>"
+            "\"3.000000000000000000\"^^<http://www.w3.org/2001/XMLSchema#float>"
```
TomyMeren, line 585: It is clear that this amount of Zeros are not necessary but do you keep it because it is how Jena returns the number?

TomyMeren:
Good (and tedious) work! I just added a comment referring to the zeros on the right side of the decimal numbers.

Do you know if this kind of combination of number types also affects arithmetic functions?

modules/engine/src/test/scala/com/gsk/kg/engine/compiler/GroupBySpec.scala
vil1: These zeroes are actually there because I convert all numbers to `DecimalType.SYSTEM_DEFAULT`. There may be a way to remove the non-significant ones, but I haven't found it yet, and I'm afraid a simple `rtrim(col, "0")` would lead to incorrect results in some cases.

TomyMeren: It makes sense, as does the use of the round method. https://github.com/apache/spark/blob/0b892a543f9ea913f961eea95a4e45f1231b9a57/sql/core/src/main/scala/org/apache/spark/sql/functions.scala#L2262
We could lose precision.

TomyMeren (approved):
LGTM!

L-Lavigne (commented):
Looks good but please see my concern about the decimal casts.

